Issue #787, opened 2014-12-10
Sort leaderboard in descending or ascending order of score


madclam:
In the AutoML challenge, the results are ordered in decreasing <Rank> instead of increasing <Rank>. 
![image](https://cloud.githubusercontent.com/assets/7594063/5384480/c482cc8a-8087-11e4-84cf-18b90aabe597.png)

Can we specify in the leaderboard in which way the results should be sorted?

PR #860, opened 2015-02-04
Fix leaderboard sorting order

ckcollab:
Related to #787


ckcollab:
- [ ] codalab.org/competitions/2321#results is not sorted on <Rank> but on another field.
- [ ] Also should be able to sort/re-sort column

PR #956, opened 2015-04-16
support arbitrary CLI commands from the action bar

percyliang:
Need to sync up with codalab-cli pull request 59.


Raisins:
Does this require some changes to the cli that are not pushed yet. I am getting an error. 
```
manager = CodaLabManager(temporary=True, clients={settings.BUNDLE_SERVICE_URL: self.client})
TypeError: __init__() got an unexpected keyword argument 'clients'
```

percyliang:
Use the 'pliang' branch for both codalab and codalab-cli

Issue #957, opened 2015-04-17
Scoring program access to all entries


madclam:
This is an issue related to:
https://github.com/codalab/codalab/issues/908

To implement coopetitions, we need to be able to compare entries with one another. In particular, one index of novelty of an entry is the number of errors that are in common with other entries: if there are few errors and there is a low overlap between these errors and other entries' errors, the new entry has a lot of added value.

To compute differences in errors made by the participants, the scoring program needs access to all the entries made, not just the user's entry.

ckcollab:
Still waiting on @lr292358 to test this

lr292358:
The reply is in the #1032

ckcollab:
Scoring program output may be lost? Try "-u" flag?

lr292358:
Adding -u didn't change too much (instead of nothing, an initial part of stdout was stored).
I added sleep 6 seconds at the end of the script and how works good, so OK.

But in general output of users codes/runtime errors before crash might not be stored, so this is still an issue --> I created a new one for it #1047, and I close this one.

ckcollab:
Thanks for the blazing fast feedback @lr292358! :)

Issue #908, opened 2015-03-11
Make it possible for Contest Organizers to see current Rankings from Scoring Program


ErickWa:
Requested by Isabelle:

* User submits new entry
* Scoring program scores entry and returns a result for output to the leaderboard

This is the current functionality.  Additional functionality required is:

* Scoring program should not only be aware of a discrete submission output, but should also have logic that compares this new output to the leaderboard, and performing an additional calculation to see how much better or worse this output is, and then award "Points" based upon the different scores.

* Leaderboard automatically re-calculates points for all entries based on relative impact of a new output.

Compare this to current functionality in which scoring program operates independently of leaderboard.

Also need to define how "points" work, and how the leaderboard changes over time

Refer to "Coopetition Points System" for more details

ErickWa:
Related to Task #909

ckcollab:
Use "Download as CSV" information and send that to scoring program

PR #984, opened 2015-04-25
Send leaderboard results to scoring program

ckcollab:
Related to issue #957 and issue #908


ckcollab:
The leaderboard results are provided to the scoring program in input/scores/metadata

PR #987, opened 2015-04-29
Edited worksheet table CSS

ppasupat:
This patch edits the CSS of the worksheet tables:

* Long cell contents are cut off but can be displayed by hovering the cursor on the cell.
* Added class names to cells to make styling easier. For example, a person with GreaseMonkey / TamperMonkey can add `.table-column-acc { color: red; }` to the personal CSS file to make the "acc" column text red.
* Column separator: If the column name is "|" (i.e., by putting `@add |` in the table schema), then a gray column separator is created

### DIFF
--- a/codalab/apps/web/static/less/worksheets.less
+++ b/codalab/apps/web/static/less/worksheets.less
@@ -265,8 +265,22 @@
                   font-size:@font-size-med;
                   border-width:2px;
                   border-color:white;
+                  white-space:nowrap;
+                  max-width:25em;
+                  overflow:hidden;
+                  text-overflow:ellipsis;
+                  &:hover {
+                     white-space:normal;
+                     overflow:auto;
+                  }
                   a.bundle-link {
                      font-weight:bold;
+                  }
+                  &.table-column-_7C {
+                     background-color:#AAA;
+                     color:#AAA;
+                     width:3px;
+                     padding:0
                   }
                }
             }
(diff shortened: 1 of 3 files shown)


codalab/apps/web/static/less/worksheets.less
```diff
@@ -54,6 +54,9 @@
          top:-7px;
       }
    }
+   > .container {
+     width:auto;
```
Raisins: This pushes all worksheets content to the left side. Off sets the page and makes long form text spread across the entire page making it very hard to read.  
left

Why change this?

Raisins: Another example of where the formatting is all off and hard to read. 
screen shot 2015-04-29 at 1 32 42 pm

ppasupat:
I see. I will revert the page alignment.
Long tables will overflow to the right though. Should I bring back the line wrap in table cells as well?

Raisins:
Maybe just an overflow should do the trick. Your current css does that on small screens and mobile screens.  Although it does a werid jump thing sometimes. I might ask @adamtlord to take a look as well.

adamtlord:
What is the goal here? @ppasupat what is the problem you'd like to solve with the way tables are displayed?

ppasupat:
I would like to compare my experiment's results in many parameter settings. I put different parameters in different columns.

Sometimes a cell contains a lot of text, which increases the row height and it is more difficult to compare different rows. And originally, the total table width is set to the text width, making the columns  very thin and tall.

adamtlord:
Ah, that makes sense. I'll work on prototyping some solutions that might address this. Thanks.

PR #1001, opened 2015-04-29
Revert the left alignment. Hide overflow only in mobile devices.

ppasupat:
Fixed the page alignment issue in #987.


Raisins:
:+1:

Issue #967, opened 2015-04-18
restrict usernames


percyliang:
Only allow usernames to be ones that match spec_util.NAME_REGEX, because people's default worksheet needs to be their username, and worksheet names cannot be arbitrary.

Raisins:
Rather then arbitrarily limit usernames across codalab. Which will be frustrating to a user. Lets strip out things that are not allowed in a worksheet. I was doing a version of this before with automatic creation of worksheets. 

https://github.com/codalab/codalab/blob/master/codalab/apps/authenz/views.py#L44

percyliang:
I think it's much better and simpler to have consistency between usernames
and worksheets.  We need to have restrictions on the username anyway (e.g.,
no spaces), so we let's not have extra logic that strips out things.  With
stripping out, you might get two unique usernames mapping to the same
worksheet name, which would be bad.

PR #1045, opened 2015-05-20
restrict usernames

Raisins:
issue #967


Raisins:
@ckcollab can you test this

ckcollab:
@Raisins how do I test this, try creating a username with things other than a-zA-Z0-9.-_?

Should add some tests! We could pair program this for sure.

Issue #1175, opened 2015-08-06
cl work autocomplete based on name


percyliang:
cl work ... should autocomplete based on name, not just uuid.

Issue #1107, opened 2015-06-12
action bar: autocomplete on bundles should work


percyliang:
Autocomplete on bundles should invoke
cl search ...
and display things nicely like cl search does.

Issue #1101, opened 2015-06-12
action bar: cl work <name> doesn't work


percyliang:
Currently, only cl work <uuid> works.
We should support cl work <name>, where <name> can be autocompleted too.

PR #1258, opened 2015-09-08
worksheet interface with new action bar to master

Raisins:


percyliang:
Note: this should not be pushed to production yet because it doesn't subsume the functionality of the old action bar (doing things by name is still missing).

Raisins:
yeah. Not ready to be pushed. Still needs name rather then UUID refactor  #1107, #1175, #1101

Issue #1237, opened 2015-08-27
Downloads empty?


andylamb:
When I try to download a bundle, e.g. https://www.codalab.org/bundles/0xdd3af6cc965e4fa8bff23fbcaf898c96/download, I get an empty file.

<img width="383" alt="screen shot 2015-08-27 at 8 56 03 am" src="https://cloud.githubusercontent.com/assets/4143035/9525548/7a2d69c2-4c99-11e5-9d44-11c86c2391af.png">

Have other people had this issue? Maybe a browser problem on my side. Tried on Safari, Chrome, Firefox.

Raisins:
Hey Andy, 
You are not alone with bug, that bundle download seems to have a bug. 
You can see the output of that bundle on this worksheet
https://www.codalab.org/worksheets/0x3229d7bfa2d24bdf97559d3b40de7588/
```
0 | price:.23 sqft:.25 age:.05 2006
1 2 'second_house | price:.18 sqft:.15 age:.35 1976
0 1 0.5 'third_house | price:.53 sqft:.32 age:.87 1924
```
Which means the file does have data, but upon downloading I am also seeing a blank file. 
Added to sprint, Will be looking into this. Please update if this happens with a different bundle.

andylamb:
It seems to happen on all bundles for me. For example, when i download https://www.codalab.org/bundles/0x96e9dcc3e0194f5aa5394a93551543dd/download, "stanford-corenlp...zip" is empty.

PR #1262, opened 2015-09-10
Fix download of bundles

percyliang:
- refactored to simplify the logic
- make more efficient by directly streaming the bundle from the service, bypassing local temp


percyliang:
Fixes #1237

PR #1272, opened 2015-09-16
Fix submission ordering

ckcollab:


ckcollab:
Reverts some changes I made for optimization that are probably unnecessary (the real problem was azure filename lookup times).

PR #1296, opened 2015-09-27
autocomplete bundle info on name if in worksheet

Raisins:
While doing `cl info ` if you start to type a name of a bundle it will autocomplete for you. Else assumes you are doing a `cl search` and just looking at any and all bundles since names repeat outside of your worksheet..  Need some discussion about what the UX flow should be. 

cl-info

issue: #1107

### DIFF
--- a/codalab/apps/api/routers.py
+++ b/codalab/apps/api/routers.py
@@ -37,6 +37,7 @@ urlpatterns += (
     url(r'^worksheets/delete/$', views.WorksheetsDeleteApi.as_view(), name='api_worksheets_delete'),
     url(r'^worksheets/search/$', views.WorksheetsSearchApi.as_view(), name='api_worksheet_search'),
     url(r'^worksheets/get_uuid/$', views.WorksheetsGetUUIDApi.as_view(), name='api_worksheet_get_uuid'),
+    url(r'^worksheets/bundle_list/$', views.WorksheetsGetBundleListApi.as_view(), name='api_worksheet_bundle_list'),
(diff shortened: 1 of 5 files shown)


percyliang:
Autocomplete for bundles shouldn't use 'cl search' because bundle names are not unique, so it's really confusing to do that.  Instead, just match on bundles in the current worksheet.  Eventually, we want this for other commands, the main one being 'cl run'.

Raisins:
So the use case is only on bundles in the current work sheet. `cl info` would not be used to find/info on other bundles? 
I kept in if you do an empty tab complete or if you start with `0x...` and hit tab it's a generic look up while if you start a name and hit tab it is only on this worksheet.

percyliang:
Yes, this is just to be consistent with the semantics of autocomplete
actually just filling out a prefix, not doing a general search given
keywords, which would be required if people actually wanted to find bundles.

We'll need a separate mechanism for search (typing in free-form keywords
and updating in real time), but that will be distinct from this
autocomplete.

Raisins:
Cool, The end points for both types of search are set up. 
Will change `cl info` to use autocomplete with only in worksheet bundles. Which can use as a model for other commands in the future.

Raisins:
Forcing worksheet on search on front end and codalab site backend. But in cli I am seeing `search -w worksheet` not respecting worksheet 
screen shot 2015-09-28 at 1 11 57 pm

percyliang:
Search shouldn't depend on the current worksheet (right now).

Raisins:
Changed to a worksheet look up similar to `cl ls`. Matches on name and only bundles in current worksheet. 
cl-info
A small bug left about name to uuid (hitting enter opens the wrong bundle info)

Issue #1201, opened 2015-08-12
Multiple Tracks (Phases) for Competitions


ErickWa:
May be applicable to Computer Vision competitions.  Evelyne to spec.

ckcollab:
Under one competition hat, have sub competitions within. For example, it is painful to make a new competition for each dataset we want to make a submission against.

ErickWa:
Hi Eric, 

The priority sounds good to me.
However, #2 and #3 are important in the long run.

We are going to host a challenge by early October.
It would be good to know the feasibility of #4 so we can decide to have multiple pages for the challenge or we can condense all challenge into one page.

Once you have prototype for #4, we can help testing it!

Best, 
Tsung-Yi

On Wed, Sep 23, 2015 at 9:35 AM, Eric Carmichael <eric.carmichael@tivix.com> wrote:
Hey Tsung-Yi,

1) Technically yes you should be able to access this information, for example if you're using Python and requests you can use a Session to auth and download things -- I believe this should be possible anyway. Making this into a documented API would be perfect but I think in its current state it should work OK?
2) (updating leaderboard with metadata/groups)
3) Multiple entries in the leaderboard can be difficult, currently a lot of things are tied to the idea/assumption of 1 submission per user on the leaderboard.
4) This one could be one of the more simple issues to solve, maybe, I will need to put more thought into this. I could try to come up with a prototype, would you be able to test it extensively? Like make a competition with phase migrations but have phases open still, that kind of stuff
5) (email notifications on submission completion)

I believe priority is in this order: #1, #4, #5.. (not sure on other two)?


Also, when might you need these things completed by? What priorities would you give each of these issues?


Thanks,
Eric




On Mon, Sep 21, 2015 at 5:14 PM, Tsung-Yi Lin <tl483@cornell.edu> wrote:
Hi Eric, 

Sorry for the late reply.
My instinct is that we can first solve easy problems, e.g., #1, #4, and #5.
It will be great to catch up later this week.  Does Friday work for you?

Below is my comments:

1) There are information that is only accessible to organizer (e.g., Download all submissions on leaderboard).  Is there any API to login CodaLab so I can access organizer-only content?

2) The detailed results feature doesn't seem to solve the problem.  The idea is that leaderboard displays more information (see reference section here for example).  The participants can update the information when they want to.  This is extremely useful for research community to share and promote works.  The current leaderboard only displays user name which is quite confusing for which group did the work and what is the corresponding paper.

[...]

PR #1326, opened 2015-10-05
Multiple active phases in a competitions

FlavioAlexander:
Resolves issue #1201

- Multiple phases open at the same time

# TODO
- [ ] Document this on wiki somewhere


ckcollab:
Code looks great so far, we'll write tests for this tomorrow!

PR #1329, opened 2015-10-10
fix cl run call from action-bar

pbhatnagar3:
This fixes cl run command in action-bar. 

Added logic to check if structured_info is present in data.data and if refs is part of data.data.structured_info.

### DIFF
--- a/codalab/apps/web/static/js/search/ws_actions.js
+++ b/codalab/apps/web/static/js/search/ws_actions.js
@@ -90,21 +90,38 @@ var WorksheetActions =  function() {
                             dataType: 'json',
                             data: JSON.stringify(postdata),
                             success: function(data, status, jqXHR){
-                                console.log('===== Output of command: ' + options);
+                                // console.log('===== Output of command: ' + options);
                                 if (data.data.exception){
                                     console.error(data.data.exception);
                                     term.echo("<span style='color:red'>Error: " + data.data.exception +"</span>", {raw: true});
                                 }
                                 if (data.data.stdout){
-                                    console.log(data.data.stdout);
+                                    // console.log(data.data.stdout);
                                     term.echo(data.data.stdout);
-                                    var references = data.data.structured_result['refs'];
-                                    Object.keys(references).forEach(function(k) {
-                                        $(".terminal-output div div:contains(" + k + ")").html(function(idx, html) {
-                                            var link = '/' + references[k]['type'] + 's/' + references[k]['uuid'];
-                                            return html.replace(k, "<a href=" + link + " target='_blank'>" + k + "</a>");
-                                        });
-                                    }, this);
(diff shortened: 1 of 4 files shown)


codalab/apps/web/static/js/search/ws_actions.js
```diff
@@ -98,13 +98,15 @@ var WorksheetActions =  function() {
...
-                                        $(".terminal-output div div:contains(" + k + ")").html(function(idx, html) {
-                                            var link = '/' + references[k]['type'] + 's/' + references[k]['uuid'];
-                                            return html.replace(k, "<a href=" + link + " target='_blank'>" + k + "</a>");
-                                        });
-                                    }, this);
+                                    if (data.data.structured_result && data.data.structured_result.refs){
+                                        var references = data.data.structured_result['refs'];
+                                        Object.keys(references).forEach(function(k) {
+                                            $(".terminal-output div div:contains(" + k + ")").html(function(idx, html) {
+                                                var link = '/' + references[k]['type'] + 's/' + references[k]['uuid'];
```
percyliang: Do we want to put some more checks here to ensure these are valid?

pbhatnagar3: I believe that the conditions cover the changes made. And I am working on writing some selenium tests so we can have a better testing framework.

percyliang: If 'type' and 'uuid' don't exist, or 'type' is not 'bundle' or 'worksheet', we have a problem...

pbhatnagar3:
Yes. I was assuming that structured_info, if available, would have the uuid and type. I have added conditions to take care if that is not the case. Also added conditions to ensure that links are created only when the type is either a bundle or a worksheet. 
 
codalab/apps/web/static/js/search/ws_actions.js
```diff
@@ -98,13 +98,18 @@ var WorksheetActions =  function() {
...
-                                    }, this);
+                                    if (data.data.structured_result && data.data.structured_result.refs){
+                                        var references = data.data.structured_result['refs'];
+                                        Object.keys(references).forEach(function(k) {
+                                            $(".terminal-output div div:contains(" + k + ")").html(function(idx, html) {
+                                                var hyperlink_info = references[k];
+                                                if (hyperlink_info.uuid && (hyperlink_info.type === 'bundle' || hyperlink_info.type === 'worksheet')) {
+                                                    var link = '/' + hyperlink_info['type'] + 's/' + hyperlink_info['uuid'];
+                                                    return html.replace(k, "<a href=" + link + " target='_blank'>" + k + "</a>");
+                                                }
```
percyliang: If there's an error, don't just fail silently!  Log it to some place and also have selenium check this.  Also, remember to fail gracefully (this is not happening now).

Issue #1273, opened 2015-09-16
render the worksheet title at the top of worksheet


Raisins:
New worksheet metadata 'title'
see PR https://github.com/codalab/codalab-cli/pull/137

PR #1330, opened 2015-10-12
#1273 render worksheet title

pbhatnagar3:
#1273 

#### worksheet with a title

werwe

#### worksheet without a title
image

### DIFF
--- a/codalab/apps/web/static/css/imports.css
+++ b/codalab/apps/web/static/css/imports.css
@@ -6427,8 +6427,10 @@ fieldset[disabled] .btn-secondary.active {
   padding-left: 50px;
 }
 .worksheet-icon {
-  text-indent: 75px;
-  background: url('/static/img/icon_worksheet.png') no-repeat 0 10px;
+  margin: 0px 0px 0px 0px;
+  display: inline-block;
+  float: left;
+  content: url('/static/img/icon_worksheet.png') no-repeat 0 10px;
   line-height: 65px;
 }
 .bundle-icon-sm {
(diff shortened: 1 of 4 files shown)


percyliang:
Could we render the title as the most prominent thing in place of the name
(a bit smaller than it is right now so we don't have too much wrapping),
and have the name underneath in a much smaller font, along with the owner?

pbhatnagar3:
Yeah sure. Just saw this. I will make those changes as soon as I am done with dinner. :)

pbhatnagar3:
The updated layout looks like the following:
new_layout

kashizui:
Looks great! Did we want attribute names (e.g. "Owner: Codalab") in the sidebar as well?

percyliang:
Looks good.  Two things:
- Make sure the title shows up in <title>...</title>
- Update the side panel to be consistent with the main page (might be even
wise to share code).  Only difference is to print out the uuid (because we
might be displaying a subworksheet).

pbhatnagar3:
Updated the side_panel too. And <title>...</title> contains the title. @percyliang @kashizui please review.

percyliang:
Is <title>...</title> actually set to worksheet.title?  Otherwise, LGTM.

pbhatnagar3:
Yup. The line ``  {% block head_title %}{{worksheet_title}} -  Worksheet details{% endblock head_title %}`` in codalab/apps/web/templates/web/worksheets/detail.html takes care of it. I double checked the source of the rendered page and it is set.

percyliang:
Ah, okay - thanks!

kashizui:
LGTM!

Issue #1324, opened 2015-10-05
stdout/stderr rendering in side panel broken


percyliang:
Currently, base64 is shown momentarily before the rendered information comes in (same for date fields like 'created').  Fix: perhaps don't render any of the fields which are going to get updated anyway.

PR #1331, opened 2015-10-12
Issue #1324 stdout stderr rendering in side panel broken

pbhatnagar3:
#1324 fixes stdout stderr rendering in side panel broken
#### When data is still being loaded
minion

#### When the data has finished loading
minion


percyliang:
We still want some state to be shown in the side panel while we're waiting
for more information (just not incorrect information like the base64
encoded stderr/stdout).  We don't want to just have the empty panel.

pbhatnagar3:

percyliang:
LGTM

Issue #1340, opened 2015-10-15
shift-g should scroll to last row of table


percyliang:
Suppose we have a worksheet with a table at the bottom.

Currently, shift-g only goes to the last table, not the last row of the table.

Issue #1269, opened 2015-09-15
Add preview when bundle is a single file


tonyyang1993:
Reference bundle:
https://www.codalab.org/bundles/0x7665d99ca3724747a10ae443cd0c90d6/

The bundle is a simple text file. At website interface, we'd better provide a preview of this file, like what we did for stdout/stderr. At command line interface, `cl cat uuid` does a great job.

percyliang:
Need to do this in both the side panel and the bundle details page.

Issue #1250, opened 2015-09-02
Unnecessary indentation of stdout/stderr @bundle details page


tonyyang1993:
In stdout/stderr preview of a run bundle, starting from the second line, there is always an extra space.

Reference links:
stdout: https://www.codalab.org/bundles/0xca09c8e2318349a2b996de9c54e2f914/
stderr: https://www.codalab.org/bundles/0x78c722b0e1994d5c91f43a293ef9ca71/

Raisins:
@tonyyang1993 saw this reopen, Update the less not the css. It gets compiled and replaced.

Issue #1067, opened 2015-05-29
side panel display file browser


percyliang:
Display file browser for bundles that are directories.
Should unify this code with the bundle details page.

PR #1342, opened 2015-10-17
Go over the entire worksheets codebase and do general clean up

percyliang:
- created file with structural overview of worksheets code (worksheets-structure.txt)
- write basic documentation for each file
- deleted unused files (e.g., old crufty non-React interface code)
- delete some of the editing code for worksheets
- minor tweaks (e.g., changing title for bundle details)

@kashizui @pbhatnagar3 Please take a look and more importantly, test it!  Hopefully this will make the worksheet code more manageable and not break anything.


codalab/apps/api/routers.py
```diff
@@ -40,12 +41,12 @@
     url(r'^worksheets/bundle_list/$', views.WorksheetsGetBundleListApi.as_view(), name='api_worksheet_bundle_list'),
     url(r'^worksheets/command/$', views.WorksheetsCommandApi.as_view(), name='api_worksheets_command'),
     url(r'^worksheets/(?P<uuid>[A-Za-z0-9]+)/$', views.WorksheetContentApi.as_view(), name='api_worksheet_content'),
+
```
kashizui: Comment for Bundles here too?

codalab/apps/api/views/worksheet_views.py
```diff
@@ -336,51 +288,42 @@ def get(self, request, uuid):
...
-                new_metadata['request_gpus'] = int(new_metadata['request_gpus']);
-                new_metadata['request_priority'] = int(new_metadata['request_priority']);
-                new_metadata['actions'] = new_metadata['actions'].split();
-                new_metadata['exitcode'] = int(new_metadata['exitcode']);
+
+                # TODO: do this generally based on the CLI specs.
+                # Remove generated fields.
+                for key in ['data_size', 'created', 'time', 'memory', 'exitcode', 'actions']:
+                    if key in new_metadata:
+                        new_metadata.pop(key)
```
kashizui: nit: `del new_metadata[key]` makes it clearer we're not using the value

codalab/apps/api/views/worksheet_views.py
```diff
@@ -465,27 +387,26 @@ def post(self, request):
...
 
-
 class BundleContentApi(views.APIView):
     """
-    Provides a web API to browse the content of a bundle.
+    Return files (map from file name to list of lines) inside the bundle.
+    If |path| is unspecified, then return stdout and stderr for a bundle.
     """
     def get(self, request, uuid, path=''):
+        maxlines = 100
```
kashizui: Maybe set `maxlines` as a static constant-ish or an optional kwarg?

codalab/apps/web/static/js/bundle/bundle_interface.js
```diff
@@ -155,15 +158,22 @@ var Bundle = React.createClass({
         var editing = this.state.editing;
         var tableClassName = 'table' + (editing ? ' editing' : '');
         var editButtonText = editing ? 'cancel' : 'edit';
-        /// ------------------------------------------------------------------
-        if(editing){
+
+        if (editing)
             saveButton = <button className="btn btn-success btn-sm" onClick={this.saveMetadata}>save</button>
```
kashizui: nit: `saveButton = (<button className="btn btn-success btn-sm" onClick={this.saveMetadata}>save</button>);`

kashizui:
This is great, very satisfying! Will be playing around with an instance of this branch on on my machine.

codalab/apps/web/static/js/bundle/bundle_interface.js
percyliang: Yes, should be semicolon.  There are still tons of style violations, which I hope we can fix all at once in another PR.

percyliang:
@kashizui: Fixed comments, made some other refactoring changes

kashizui:
I can't seem to click back to View after going into View Source? Not sure if this is related to this patch

percyliang:
Fixes #1340 #1269 #1250 #1067

percyliang:
@kashizui @pbhatnagar3 : made another pass and cleaned up the react logic; bugs hopefully should be ironed out now, please take another look.

kashizui:
Running a local server, seems to be working! I say LGTM but @pbhatnagar3 may have more to say about the ReactJS side of things.

codalab/apps/web/static/js/worksheet/worksheet_interface.js
```diff
@@ -148,28 +152,26 @@ var Worksheet = React.createClass({
...
-        this.setState({updating:true});
+        this.setState({updating: true});
         ws_obj.fetch({
-            success: function(data){
-                // console.log("fetch_and_update success");
-                // console.log("%c--------------------------------------------------------------", "color: Green; font-size:15px;");
-                if(this.isMounted()){
+            success: function(data) {
+                if (this.isMounted()) {
                     this.refs.list.setState({worksheet: ws_obj.getState()});
```
pbhatnagar3: This may be an anti pattern for React. If we are setting the state of anything other than self in a component's class, that indicates that the source data is not stored at one place and we have multiple copies which is not ideal when the virtual DOM is created and the diffing happens. 

However, to fix this, we would have to do some major re-factoring of the code.

percyliang: Yes, this is definitely not good, and I've been trying to fix this so that
they are more React-like.  Updated the code with a TODO.  Let's get this PR
checked in soon (it's already grown a bit out of control), so we can build
on top of it.

pbhatnagar3: The changes look good. :+1: 

One comment on the React way of doing things though. Technically, since React creates a virtual DOM to figure out which part of the actual DOM need to be updated, all the JQuery code should not be a part of our component class. Components should just be accessing the data passed in the parent in `this.props` and changing its state by modifying `this.state`. For all the JQuery commands, we should ideally just have the logic and the variables checking things need to be shown or hidden and when the data would change, the logic described by the components render and ComponentDidMount method would help create the virtualDOM that would update the div state itself. But that would require major restructuring of code and creation of more components that handle the error and other JQuery logic.

percyliang: Agreed - we should probably rely less on jquery, which is an easy crutch.

pbhatnagar3: Also, the AJAX call in line 231 of worksheet_sidepanel sees to be giving error 500 (internal server error) when user clicks on a bundle that is a dataset and does not have any associated sub-files with it. Here is a screenshot: 
error

Should we handle this at the cli side?

percyliang: I'm not able to reproduce this...do you have the latest CLI, and can you be
more specific about what you tried and what error you got?

pbhatnagar3: Yeah sure. I am using the following worksheet for testing. 

```
// Editing worksheet links(0xe87e3769a23947d38e58bdebbc28c068).
// https://github.com/codalab/codalab-cli/worksheet-syntax.md
//
[run run-echo-hello : echo hello]{0x3ba845bc288d466aa0d2e846b0242c95}
[dataset Pujun_Bhatnagar_Resume.pdf]{0x59103debd7744d8d899c5c666a0a82e8}
```

where the first item is a run while the second one is just a dataset. When I open the worksheet on the web-interface. The following is the activity of the console. (there were a bunch of warnings from React but I cleared the console to see what's exactly happening)


^ till now no error. I click on the first bundle and see the associated sub-files. 


^ still no error and the stdout and stderr are shown in the side panel. 
However, when I click on the second bundle, which is a data set, I get an internal 500 error. 

This is happening in line 231 in worksheet_sidepanel.js 
I uploaded another blank file `Pikachu.txt` as a dataset and that does not give this error.


However, when I am trying to do the same with .pdf files, this error happens. 

This is definitely odd. Might be related to pdf files. I was just playing around with this and came across this.

percyliang: I see, it's only an issue with binary files, which I think we're trying to
treat as utf8.  Do you know what's the easiest way to fix this?

kashizui:
🎉

PR #1345, opened 2015-10-19
1335 homepage improved

FlavioAlexander:
- Made some adjustments for the new homepage

### DIFF
--- a/codalab/apps/common/utils.py
+++ b/codalab/apps/common/utils.py
@@ -5,15 +5,14 @@ from django.conf import settings
 from apps.web.bundles import BundleService
 
 
-def random_worksheets(request_user, limit=3):
-    """Function to return most random worksheets, if any"""
+def get_worksheets(request_user, worksheet_uuids, limit=3):
     service = BundleService(request_user)
     list_worksheets = service.worksheets()
 
     if not list_worksheets:
         return list_worksheets  # just incase the list is empty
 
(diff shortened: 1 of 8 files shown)


codalab/apps/common/utils.py
```diff
@@ -1,15 +1,18 @@
...
-    if not worksheets:
-    	return worksheets  # just incase the list is empty
+from django.conf import settings
 
-    sorted_worksheets = sorted(worksheets, key=lambda k: k['id'], reverse=True)
 
-    if len(sorted_worksheets) <= 2:
-        worksheets = [(val['uuid'], val['name'], val['owner_name'])for val in sorted_worksheets]
-        return worksheets
+def random_worksheets(list_worksheets):
```
percyliang: Getting the list of all worksheets and then filtering is expensive.  Once we have worksheet tags, we should be using search_worksheets to do that.  For now, put all the logic (getting the list of worksheets and filtering) in one place (here) so we can change it later easily.

percyliang:
Also note that there will be a merge conflict with #1342

FlavioAlexander:
@percyliang I agree with "Getting the list of all worksheets and then filtering is expensive" Let's just wait for the template tags and then we should able to change the logic. Will put all logic under utils.py so it's easy to make changes later in the future. Thank you for pointing out the merge conflict :+1:

PR #1349, opened 2015-10-21
Selenium testing

pbhatnagar3:
#### Selenium Testing

Here is what I am doing for the frontend testing. I had to modify some of them to take #1342 into account but these should work now. :) Please let me know if you have any suggestions. I will continue to write more. 

@percyliang @kashizui This is not ready to be merged. Still a work in progress.

### DIFF
--- a/tests/TEMP_README.txt
+++ b/tests/TEMP_README.txt
@@ -1,0 +1,17 @@
+Tests for codalab are written in Selenium Driver
+
+INSTALLATION
+
+For installing selenium, run the following while inside your virtualenv
+
+	pip install -U selenium
+
+For Google Chrome support:
+	
+	1. download the ChromeDriver from https://code.google.com/p/selenium/wiki/ChromeDriver
+	2. move it to /usr/local/bin
+	3. set appropriate permissions
+		chmod 777 /usr/local/bin/chromedriver
+
+
+
--- a/tests/TEMP_README.txt
+++ b/tests/TEMP_README.txt
@@ -4,6 +4,7 @@ INSTALLATION
 
 For installing selenium, run the following while inside your virtualenv
 
+	source venv/bin/activate
 	pip install -U selenium
 
 For Google Chrome support:
--- a/tests/test.py
+++ b/tests/test.py
@@ -1,24 +1,50 @@
 import unittest
 from selenium import webdriver
 from selenium.webdriver.common.keys import Keys
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.ui import Select
+from selenium.common.exceptions import NoSuchElementException
 
 class CodalabTestCase(unittest.TestCase):
 
     def setUp(self):
+        print "blah"
         self.browser = webdriver.Firefox()
         self.addCleanup(self.browser.quit)
         self.base_url = "http://localhost:8000/"
 
-    def testPageTitle(self):
+    def testBaseURL(self):
         self.browser.get(self.base_url)
+        self.browser.maximize_window()
         self.assertIn('CodaLab - Home', self.browser.title)
-
-    def testWorksheetCompetionbuttons(self):
+        for i, item in enumerate(['Worksheets', 'Competitions', 'Help']):
+            self.assertEqual(
+                item,
+                self.browser.find_element_by_xpath("html/body/nav/div/div[2]/ul/li[" + str(i+1) + "]/a").get_attribute('innerHTML')
+            )
+        for i, item in enumerate(['Create an Experiment!', 'Join a challenge!']):
+            self.assertEqual(
+                item,
+                self.browser.find_element_by_xpath("html/body/div[2]/div/div/div[" + str(i+1) + "]/div/div/a").get_attribute('innerHTML')
+            )
+        
+        assert "Run reproducible experiments and created executable papers using worksheets" in self.browser.page_source
+        assert "Enter the existing competition to solve challenging data problems, or host your own." in self.browser.page_source
+    
+    def testWorksheet(self):
         self.browser.get(self.base_url)
-        experiment_btn = self.browser.find_element_by_link_text("Create an Experiment!")
-        self.assertIsNotNone(experiment_btn, msg=None)
-        challenge_btn = self.browser.find_element_by_link_text("Join a challenge!")
-        self.assertIsNotNone(challenge_btn, msg=None)
+        self.browser.find_element_by_xpath("html/body/nav/div/div[2]/ul/li[1]/a").click()
+        self.assertEqual('Worksheets',
+            self.browser.find_element_by_xpath("html/body/div[1]/div/h1").get_attribute('innerHTML')
+        )
+        
+        
+    # def testWorksheetCompetionbuttons(self):
+    #     self.browser.get(self.base_url)
+    #     experiment_btn = self.browser.find_element_by_link_text("Create an Experiment!")
+    #     self.assertIsNotNone(experiment_btn, msg=None)
+    #     challenge_btn = self.browser.find_element_by_link_text("Join a challenge!")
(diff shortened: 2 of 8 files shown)


tests/TEMP_README.txt
```diff
@@ -0,0 +1,18 @@
+Tests for codalab are written in Selenium Driver
```
percyliang: Why is this called TEMP_README?

tests/TEMP_README.txt
```diff
@@ -0,0 +1,18 @@
...
+
+INSTALLATION
+
+For installing selenium, run the following while inside your virtualenv
+
+	source venv/bin/activate
+	pip install -U selenium
```
percyliang: replace 2 lines with venv/bin/pip -U selenium

percyliang:
LGTM

tests/TEMP_README.txt
pbhatnagar3: Oh. I just called it temp because I thought that we might move these instructions to some other document.

percyliang: Let's just call it README - we can still move things later.

pbhatnagar3:
Have added more tests for the worksheet. There is a lot that can be done for testing so I would keep on adding tests whenever I get time.

For now, I am using time to make sure that tests are getting enough time to render all the result of AJAX calls. This is definitely not ideal and I am trying to find an elegant way of doing this. A lot of selenium users have faced this and have suggested different solutions for diffferent use cases. Take a look at the following if you get time:  http://www.obeythetestinggoat.com/how-to-get-selenium-to-wait-for-page-load-after-a-click.html But for the meantime, these tests should work.  

@percyliang @kashizui I think this is now ready to be merged in. Please take a look and if possible, run them on your local machine if you get time.

tests/test.py
```diff
@@ -0,0 +1,204 @@
...
+        self.base_url = "http://localhost:8000/"
+
+    def testBaseURL(self):
+        self.browser.get(self.base_url)
+        self.browser.maximize_window()
+
+        # check title
+        self.assertIn('CodaLab - Home', self.browser.title)
+
+        #check the list in the top right
```
percyliang: nit: space

tests/test.py
```diff
@@ -0,0 +1,204 @@
...
+                self.browser.find_element_by_xpath("html/body/nav/div/div[2]/ul/li[" + str(i) + "]/a").get_attribute('innerHTML')
+            )
+
+        for i, item in [(1, 'Worksheets'), (2, 'Competitions')]:
+            self.assertEqual(
+                item,
+                self.browser.find_element_by_xpath("html/body/div[2]/div/div/div[" + str(i) + "]/div/div/a").get_attribute('innerHTML')
+            )
+
+        assert "Run reproducible experiments and created executable papers using worksheets" in self.browser.page_source
```
percyliang: These things (static content) are not necessary to test - they will show up probably anyway and will probably change, so I'd remove.

tests/test.py
```diff
@@ -0,0 +1,204 @@
...
+        assert "Enter an existing competition to solve challenging data problems, or host your own." in self.browser.page_source
+
+        for i, item in ([(3, 'Example Worksheets'), (4, 'Latest Competitions')]):
```
percyliang: This seems very sensitive to the actual layout of the page...again, I'd make the test a bit more robust to changes in wording, ordering...  The main thing is to make sure we don't get a 404 for example worksheets.

tests/test.py
```diff
@@ -0,0 +1,204 @@
...
+            self.assertEqual(
+                item,
+                self.browser.find_element_by_xpath("html/body/div[2]/div/div/div[" + str(i) + "]/h2").get_attribute('innerHTML')
+            )
+
+    def testWorksheet(self):
+        self.browser.get(self.base_url)
```
percyliang: This should be base_url + '/list'
since '/worksheets' can also go to a default worksheet.

tests/test.py
```diff
@@ -0,0 +1,204 @@
...
+    def testWorksheetContent(self):
+        cl = 'cl'
+        self.browser.get(self.base_url)
+        self.browser.maximize_window()
+        
+        # create a temp worksheet for testing
+        temp_worksheet_name = 'test-codalab-' + str(random.randint(0, 1000000))
+        temp_worksheet_title =  'The really long testing title for testing the automated testing of worksheets'
+        temp_worksheet_uuid = self.run_command([cl, 'new', temp_worksheet_name])
+        temp_file_name = 'cl-web-interface-testing.txt'
```
percyliang: This needs to be deleted.

tests/test.py
```diff
@@ -0,0 +1,204 @@
...
+        temp_worksheet_paragraph_content = "For running automated testing on the cli " * 10
+        self.run_command([cl, 'work', temp_worksheet_uuid])
```
percyliang: It'd be better if you were running these commands through the action bar on the website...is this easy to do?

percyliang:
My main comment is that I think we shouldn't test the exact layout (we shouldn't have hardcoded long strings of divs with exact indices) since these things can change.  We just want partial tests like - the worksheet uuid appears *somewhere* on the page without regards to formatting.  Even if we were clicking on things and making sure things don't crash, that would already be a big step up.

For now, let's just not make this depend on the CLI (if possible), reduce the hardcoding, and then just check it in.

pbhatnagar3:

tests/test.py
```diff
@@ -0,0 +1,106 @@
...
+        self.run_command([cl, 'wedit', '-t', temp_worksheet_title])
+        self.run_command([cl, 'add',  '-m', temp_worksheet_paragraph_content])
+        self.run_command(['touch', temp_file_name])
+        temp_file_uuid = self.run_command([cl, 'upload', 'dataset', temp_file_name])
+
+        # click the worksheet button on the top right of the main page
+        self.browser.find_element_by_xpath("html/body/nav/div/div[2]/ul/li[1]/a").click()
```
percyliang: Too hard coded - can you search for 'Worksheets' instead?

tests/test.py
```diff
@@ -0,0 +1,106 @@
...
+        # delete the temp-worksheet
+        self.run_command([cl, 'wrm', '--force', temp_worksheet_name])
+
+        self.run_command(['rm', temp_file_name])
+
+    def worksheetContentCheckHelper(self, temp_file_name, temp_worksheet_uuid, temp_worksheet_paragraph_content):
+        # test paragraph content inserted at top of temp_worksheet
+        self.assertIn(temp_worksheet_paragraph_content, self.browser.page_source)
+        # check table in worksheet
+        table_content = ['uuid','name', temp_file_name, 'description', 'bundle_type', 'dataset', 'created', 'dependencies', 'command', 'MISSING', 'data_size', '0','state', 'ready']
```
percyliang: We don't have 'MISSING' anymore, right?  Does this check still pass?

percyliang:
Please merge into develop, not master.
Also, some of the comments about using the action bar are not addressed still.

tests/test.py
pbhatnagar3: oh no. I know what is going on. The branch had not been updated and it still shows MISSING. I did not realize that. I will pull the latest code and revise this. Sorry my bad.

tests/test.py
pbhatnagar3: yup. But that would break if there are multiple instances of word 'Worksheets'. That is why I was doing it this way. I will make the changes.

pbhatnagar3:
@percyliang @kashizui I have to close this pull request since this is requesting merge to master and not to dev. I cannot find a way to change that. I will create another pull request.

pbhatnagar3:
here is the link to the new pull request. https://github.com/codalab/codalab/pull/1382

Issue #1323, opened 2015-10-05
allow editing of bundle metadata in the side panel


percyliang:
Be able to edit the bundle name, description, tags.  Display these first, and don't allow editing of generated fields.  Should be able to click on metadata field (edit icon next to it that shows up when mouseover the field or something) and it should just show a textbox for that particular field, not everything.

PR #1359, opened 2015-10-26
allow editing of info in side-panel

pbhatnagar3:
#1323 enhancement: Allows editing of the worksheet title

#### Testing

For testing, I have taken videos because it is harder to show the behavior otherwise. Here are the links:
https://db.tt/0Wc0ae8E
https://db.tt/1nIwbj7Z


codalab/apps/api/views/worksheet_views.py
```diff
@@ -298,9 +298,9 @@ def post(self, request, uuid):
 
                 # Convert to ints
                 for key in ['request_cpus', 'request_gpus', 'request_priority']:
```
percyliang: Given that you're now getting the metadata specs, you shouldn't need to hard code any of the metadata keys here - we should just be able to get the type.

percyliang:
This is great!
Question: what happens when the user tries to type something invalid (invalid name with spaces for instance)?  Where does the error message show up?

In the future, we want to be able to support other forms of input (not just free text), for example, for tags.  Good enough for now though.

codalab/apps/api/views/worksheet_views.py
pbhatnagar3: Yup yup. I will make this change. I was unsure if this was used anywhere else in the code and I decided not to play with this earlier.

pbhatnagar3:
if the user puts in something that is invalid (like the name field),   codalab-cli/codalab/client/remote_bundle_client.py", line 148, in inner
    raise UsageError(e.faultString[index + 1:]) raises an error and the name is not updated. However, the name on the worksheet was temporarily changed to a invalid string. 


I have addressed the issue. @percyliang , please see the following screenshot (they show the updated behavior):

^ in the following, I had manually set the permission to edit as False to test things out 


^ The following is the use case when the name contains a space. 

In case of an error, the value does not get updated and it returns to the original (even in the sidepanel) when the user clicks out of the textbox.

Thanks for pointing this out. :)

I have tried using the tools and writing the code such that more than text implementation can be easily extended.  For instance, this can be done:


^ The tags use case

percyliang:
Great, thanks!  LGTM

PR #1384, opened 2015-11-07
enable side-panel editing of name and title and box-effect for editable field

### LINKED ISSUE #1380: editing in side panel broken
1) Need to be able to edit the name of a bundle on the side panel.

2) When click to edit the 'name' field, it makes me edit the 'memory' field instead. 

3) Cursor should change from pointer to the text input, not the hand when mouse over an editable field.

pbhatnagar3:
fixes #1380 
#### Testing
All the selenium tests pass

#### Manual Testing
roundededges
roundededges1

NOTE: I was unable to recreate the issue with editing the memory but I will keep on looking into it. 


percyliang:
LGTM

PR #1389, opened 2015-11-08
ace editor

pbhatnagar3:
@percyliang @kashizui Here is a pull request that is associated with ace editor. This is still a work in progress. Not ready to merge.

I have made modifications to make sure that this works with our react app and have tested out the functionality. Till now, everything is working (the codalab-secific syntax highlighting does has not been implemented but the normal syntax highlighting works)

Here is a image preview:
m

and when the user presses the 'e' key or clicks the edit button, we get the following:
m1

here is a link to the video if you want to see it working
https://www.dropbox.com/s/flunuyjg482yira/ace-editor.mp4?dl=0


percyliang:
Cool!
- Can we rename src-noconflict to ace-editor?
- Saving doesn't work now (javascript error).
- Pressing 'e' doesn't set focus in the right place.
- If the markdown gets long, scrolling is messy - there's a scrollbar on the pane but also scrolling in the editor.  This needs to be made smoother.
- Looking ahead, is it easy to have things like autocomplete / autosuggest?

pbhatnagar3:
yeah I will take care of the renaming. 
For the saving and other features like focusing on the right place, I will be working on those changes and should be done by today.
 I did not test out the scrolling issues. I will look into that. Thanks for the heads up. :+1: 
judging from the APIs that ace offer, I think that it should not be heard to implement the auto complete and the auto-suggest but I have not tried that yet. 
Ace offers really nice APIs and has really nice infrastructure that we could take advantage to implement the features that we want for editing the worksheet. Working on this right now. will keep you guys updated. :)

percyliang:
Cool, sounds good.  Asking about autocomplete not because we need it
immediately, but because we want to be able to support it eventually.  More
important to get something simple working well quickly.

pbhatnagar3:
@kashizui @percyliang This is ready to review. I have tested it and everything seems to work. I think we should merge this in. This pull request is already getting super big and I can continue working on it to implement the navigation part of the editor.

percyliang:
Is there a way not to check in all this 3rd party code?  For example, with
MathJax, we either just load it from mathjax's website or we use ./manage
prep_for_offline, which downloads the source from the server and unpacks it.

pbhatnagar3:
I am not sure. But I can take a look into it tonight.

kashizui:
New PR #1399 that doesn't include ace editor code.

PR #1394, opened 2015-11-10
Feature/worksheet content

percyliang:
Major cleanup of worksheet content.
    
    - Get rid of state, use props when possible.
    - Remove indirection (item.state => item).
    - Remove cruft in WorksheetContent regarding reverse engineering alignments.
    - Better error handling when worksheet fails (still load things).
    - Remove checkboxes.

@pbhatnagar3 please review


codalab/apps/web/static/js/worksheet/worksheet_items.js
```diff
@@ -82,11 +80,12 @@ var WorksheetItemList = React.createClass({
         this.props.updateWorksheetFocusIndex(index);  // Notify parent of selection (so we can show the right thing on the side panel)
         if (index != -1 && subIndex != null) {
           // Change subindex (for tables)
-          var item = this.state.worksheet.items[index].state;
-          if (item.mode == 'table') {
-            if (subIndex == 'end') {
-              subIndex = item.bundle_info.length - 1;  // Last row of table
-            }
+          var item = info.items[index];
+          console.log('EEEEEEEEEEE', item);
```
pbhatnagar3: remove

pbhatnagar3:
Changes looks good. Since major chunk of the worksheet content is now gone, I am not entirely certain if the updating of the various contents (in the case where you are editing things on the worksheet) will work in the same way. i am going to checkout this branch and test this.

percyliang:
It should be fine because things are now getting passed properly down using
props instead of using state.  But go ahead and test it anyway (remember to
get the latest version since I made a few updates).

pbhatnagar3:
yup yup. Everything works without any console errors. This is soooo good :+1:

PR #1414, opened 2015-11-18
'u' pastes uuid and 'a' pastes args into console.

### LINKED ISSUE #1376: type shortcut to copy bundle uuid into action bar
Use case: enter 'cl mimic' in action bar.
Then scroll around to the different bundles, and type 't' to copy the uuid into the action bar.
This makes it a lot eaiser to create new commands.
Should be able to type 't' from action bar too.

kashizui:
- `u` pastes uuid of focused bundle or worksheet into console
 - `a` pastes args of focused bundle if available
 - consolidated some key binding code (decided to go this route because alternative was the pass `activeComponent` as an additional prop to each of the table rows.

(apologies for potentially hacky looking code, in general really itching to make all this code organized better but I don't even know where to start. perhaps from the leaf nodes of the react component graph.)

Fixes #1376 

### DIFF
--- a/codalab/apps/web/static/js/worksheet/table_bundle_interface.js
+++ b/codalab/apps/web/static/js/worksheet/table_bundle_interface.js
@@ -46,8 +46,8 @@ var TableBundle = React.createClass({
         // Paste args of focused bundle into console
         Mousetrap.bind(['a'], function(e) {
             var bundleInfo = this.refs['row' + this.state.rowFocusIndex].props.bundleInfo;
-            if (bundleInfo.command != null) {
-                $('#command_line').terminal().insert(bundleInfo.command);
+            if (bundleInfo.args != null) {
+                $('#command_line').terminal().insert(bundleInfo.args);
             }
         }.bind(this), 'keydown');
     },
--- a/codalab/apps/web/static/js/worksheet/table_bundle_interface.js
+++ b/codalab/apps/web/static/js/worksheet/table_bundle_interface.js
@@ -32,11 +32,20 @@ var TableBundle = React.createClass({
             else
               this.focusOnRow(newIndex);
         }.bind(this), 'keydown');
+
+        Mousetrap.bind(['enter'], function(e) {
+            window.open(this.refs['row' + this.state.rowFocusIndex].props.url, '_blank');
+        }.bind(this), 'keydown');
+
+        // Paste uuid of focused bundle into console
+        Mousetrap.bind(['t'], function(e) {
+            var uuid = this.refs['row' + this.state.rowFocusIndex].props.uuid;
+            $('#command_line').terminal().insert(uuid);
+        }.bind(this), 'keydown');
     },
 
     scrollToRow: function(index) {
         var __innerScrollToRow = function(index) {
-            //console.log('__innerScrollToRow', index);
             // Compute the current position of the focused row.
             var node = this.getDOMNode();
             var nodePos = node.getBoundingClientRect();
@@ -64,8 +73,6 @@ var TableBundle = React.createClass({
     },
 
     render: function() {
-        //console.log('TableBundle.render', this.state.rowFocusIndex);
-
         if (this.props.active && this.props.focused)
           this.capture_keys();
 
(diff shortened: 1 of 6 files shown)


codalab/apps/web/static/js/worksheet/table_bundle_interface.js
```diff
@@ -32,11 +32,28 @@ var TableBundle = React.createClass({
...
+        Mousetrap.bind(['u'], function(e) {
+            var uuid = this.refs['row' + this.state.rowFocusIndex].props.uuid;
+            $('#command_line').terminal().insert(uuid);
+        }.bind(this), 'keydown');
+
+        // Paste args of focused bundle into console
+        Mousetrap.bind(['a'], function(e) {
+            var bundleInfo = this.refs['row' + this.state.rowFocusIndex].props.bundleInfo;
+            if (bundleInfo.command != null) {
+                $('#command_line').terminal().insert(bundleInfo.command);
```
percyliang: use args instead of command to be consistent with CLI.  command is something more specific.

codalab/apps/web/static/js/worksheet/table_bundle_interface.js
```diff
@@ -32,11 +32,28 @@ var TableBundle = React.createClass({
             else
               this.focusOnRow(newIndex);
         }.bind(this), 'keydown');
+
+        Mousetrap.bind(['enter'], function(e) {
```
percyliang: Just curious - why move the enter here?  It's fine with me.

codalab/apps/web/templates/web/worksheets/detail.html
```diff
@@ -70,6 +70,12 @@
                       <td><kbd>enter</kbd></td><td>Open current bundle or worksheet</td>
                     </tr>
                     <tr>
+                        <td><kbd>u</kbd></td><td>Paste UUID of current bundle or worksheet into console</td>
+                    </tr>
+                    <tr>
+                        <td><kbd>a</kbd></td><td>Paste run arguments of current bundle into console</td>
```
percyliang: not just run arguments - because it works for make bundles too.  Replace with 'paste arguments to recreate current bundle'.

Also, we should be consistent, and call 'console' the 'web terminal'.  Ideally, we would rename 'action bar' (it was a bar at some point), but not for this PR....

codalab/apps/web/static/js/worksheet/table_bundle_interface.js
kashizui: The handler-setting code in the original location was actually being called regardless of whether the action bar was active or not. Decided to go this route because alternative was to pass `activeComponent` as an additional prop to each of the table rows.

codalab/apps/web/static/js/worksheet/table_bundle_interface.js
kashizui: This looks pretty hard to pull out:
https://github.com/kashizui/codalab-cli/blob/7af1516f63a149480996d5c3e93fee541ab6cc9c/codalab/lib/worksheet_util.py#L354-L354

What do you think would be the best strategy here?

percyliang: We should just compute it on the webserver side and pass it into the client.

kashizui: OK, through an AJAX call to the client? Or maybe some hack through the web terminal?

percyliang: We should stick it into bundle_info before shipping it off to the client to
avoid an extra server call.

kashizui: Beautiful.

percyliang:
This is great.
Should we set focus to the web terminal after pasting the uuid/args in?  I think it's very likely that the user will need to switch back into it the web terminal (to hit space or enter, etc.)

kashizui:
I can do that tomorrow---meanwhile yes, I certainly don't see a reason why would not want to jump back to the terminal, unless you want to post multiple uuids. In which case we'll have to do some fancy space-adding as well.

kashizui:
Focuses on action bar after successful paste.

@percyliang

percyliang:
LGTM!

PR #1427, opened 2015-11-22
Escape HTML in contents bundle

### LINKED ISSUE #1423: display contents directive with binary contents crashes app

kashizui:
Let React escape any content that gets displayed in a `% display contents` bundle.

Helps fix #1423 

@percyliang please review


percyliang:
LGTM

Issue #1369, opened 2015-10-30
title modification breaks when not authenticated


pratiksha:
When I am not signed in, the CodaLab web interface allows me to click on the worksheet title and modify it in the text box. When I click outside the text box to save the title, an error message appears underneath the text box in large red letters:
````
"<Fault 1: \"<class 'codalab.common.AuthorizationError'>:Not authenticated\">"
````
I'm guessing the right behavior is to not allow a user without worksheet write permissions to open the title text box at all.

PR #1459, opened 2015-12-02
check authentication before editing title and name

### LINKED ISSUE #1429: Field-editing not handling some authorization errors
Unauthenticated users can attempt to "edit" the title of the worksheet, without displaying the authorization error that the server responds with.

pbhatnagar3:
fixes #1429 #1369 
 
Earlier, since the code checked for response.error and not response.exception, there was no error being thrown. When that was changed, the error that was thrown looked non-ideal.

- made changes to the css to make sure that the error looks more readable now. 

This was done as a precautionary measure that if ever, while editing the name or the title, an error happens, then the user should be able to read it.

also added a authentication check before allowing user to edit the name/title of worksheet. If the user has permission, worksheet looks like this:

else the worksheet looks like this (the title is no longer allowed to be edited)

### DIFF
--- a/codalab/apps/web/static/css/imports.css
+++ b/codalab/apps/web/static/css/imports.css
@@ -8780,8 +8780,11 @@ fieldset[disabled] #worksheet-list .worksheet-tile .worksheet-inner button.delet
 }
 .editable-field:hover {
     cursor: text;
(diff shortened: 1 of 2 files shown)


percyliang:
LGTM

Issue #1460, opened 2015-12-02
view source => gray out don't edit


pbhatnagar3:
show the source but do not allow the user to edit or safe any changes.

percyliang:
Would be nice to show a gray background.

FlavioAlexander:
@pbhatnagar3 I believe this ticket is part of worksheets. If yes, can I close it?

PR #1461, opened 2015-12-03
ace editor readonly when no permissions to edit

pbhatnagar3:
#1460 
Earlier, even when the user did not have permissions to edit the worksheet, they could go into the editing mode and start making changes (although they were not allowed to make safe those changes). Now the editor goes into the 'readonly' mode where they can see the raw worksheet but cannot make any edits. 

NOTE: When the user does not have permissions, the safe button is 'disabled'. However, the way CSS is written for group-btn is making it nearly impossible to make the safe button have gray background. (see the screenshot). I spent some time trying to fix this but could not. I will tackle other issues first and then come back to this CSS issue.

y_u_css

UPDATE: 
The css is fixed now. I have gutted out the css that was duplicate and was causing issues:
css_works


percyliang:
LGTM

PR #1470, opened 2015-12-04
Put CSS Hacks and local modifications into Less files

ppasupat:
The hacked modification in `imports.css` and the local modifications found in the current main web site <https://worksheets.codalab.org/static/css/imports.css> are put into Less files.


pbhatnagar3:
@ppasupat You are a genius. :+1: Thanks for this. :dancer:

PR #1476, opened 2015-12-06
Update URL dynamically when switching worksheets

### LINKED ISSUE #1457: cl work worksheet without reloading entire page (need to change URL)

kashizui:
With support for using the back/forward buttons. Things move much faster now!

Fixes #1457 


percyliang:
Code looks good to me.
There's a bug where if you switch to another worksheet and try to edit the title, it shows the old title.
We should test this carefully to make sure we don't introduce weird bugs.

kashizui:
That is a vexingly difficult bug to figure out..

kashizui:
I think this bug is related to the one where it seems like we're editing the wrong metadata field in the side bar.

kashizui:
Fixed the bug..

@percyliang please review

codalab/apps/web/static/js/worksheet/editable_field.js
```diff
@@ -0,0 +1,47 @@
+/** @jsx React.DOM */
+
+var EditableField = React.createClass({
```
percyliang: Thanks for doing this!

percyliang:
LGTM

PR #1493, opened 2015-12-09
Fix metadata field editing

kashizui:
Use the new EditableField node, which fixes all the funkiness with field editing.

Also deselect upload button after upload.


pbhatnagar3:
LGTM

kashizui:
Also only send updated metadata, which was causing problems with value validation on the server-side (editing `description` would raise error for editing `time_user` because the formatted string value of the that field was being sent along with the metadata updates).

percyliang:
LGTM!

PR #1502, opened 2015-12-12
Added PIP requirements for OS X

andreweduffy:
There's a bug in the old version of uwsgi that prevents building on
OS X 10.11, see https://github.com/unbit/uwsgi/issues/1056. Fix is to have a
requirements file with an updated version for OS X.

### DIFF
--- a/codalab/requirements/dev_osx.txt
+++ b/codalab/requirements/dev_osx.txt
@@ -1,0 +1,8 @@
+-r common.txt
+supervisor==3.0
+uWSGI==2.0.11.2
+pycrypto==2.6
+MySQL-python
+# Use sckoo@stanford.edu's patched argcomplete, until pull request has been accepted
+# https://github.com/kislyuk/argcomplete/pull/118
+git+https://github.com/kashizui/argcomplete.git@allow-overridden-subparsersaction


kashizui:
Thanks for finding this!

If this new version of uWSGI works fine, do you see any problem with simply upping the version for everyone in the `dev_azure_nix.txt` requirements file? I realize it is a different major release but if it works, that would make the requirements easier to maintain. (I can look into this too if you're not sure.)

cc: @percyliang

andreweduffy:
I haven't done any testing and I'm still getting my bearings with the project so not certain. Agree that unified requirements would be easier to maintain though. Also it seems like uWSGI went 2.0 two years ago so using up-to-date dependencies is also nice :)

kashizui:
I agree—we'll do some testing and if uWSGI 2.0 works without problem, we'll plan to just update the *nix requirements. Thanks again!

kashizui:
Let's go ahead and just up the version for uWSGI. I don't think we use any of uWSGI's specific features, and production server uses Nginx anyways. We'll deal with issues as they come up.

Closing this and filing a pull request on codalab-worksheets.

PR #1514, opened 2016-01-22
Hide scrollbars if not used on list

emersion:
This is SO ugly :P

screen shot 2016-01-22 at 16 44 30


FlavioAlexander:
@emersion :( I wish I was a designer

PR #1601, opened 2016-10-06
Update doc + Docker Setup

lsenta:
This patch provides:
- A simple way to start a MySQL server using Docker
  - `docker-compose up`, ready to go.
- Updated documentation,
  - Many parts of the setup documentation are unclear, deprecated and/or spread around multiple repositories, try to merge most of them.
- Documentation as part of the repository:
  - Makes it possible to reference parts of it and create Pull Requests,
  - Easier to maintain, code documented is in the same place than the documentation itself.
- Drop code that is not required anymore ("documented out").

We should cover and be able to drop:
- https://github.com/codalab/codalab-competitions/wiki/Setup-Local-Competitions#user-content-get-the-source-code
- https://github.com/codalab/codalab-competitions/wiki/Dev_Set-up-data-storage
- https://bitbucket.org/percyliang/codalab-deployment/src/691bb8302553624ae22d84d416faa5781ca465d5/competitions/docs/setup_azure.md?at=master&fileviewer=file-view-default


lsenta:
@madclam @ErickWa @FlavioAlexander
Can you take a look at the dev doc and let me know if you can follow the process along?
https://github.com/codalab/codalab-competitions/blob/eb69bf3bfb026eaba904848c7d0539c7a9ea4573/docs/DEV.md

FlavioAlexander:
@lsenta I will look into this this week. Looking good so far 👍

FlavioAlexander:
@lsenta Going to fix the conflicts so can merge it. We should keep `.sh` and `.bat` files though. When setting up codalab locally, those files will do the a lot of the work for us.

lsenta:
@FlavioAlexander 
I'll leave it up to you:

IMO we don't need to maintain this code because it can be replaced by:
"requires python2.7. Install the xxx/requirements.txt in a virtualenv".

The Python community has a lot of documentation about this
and there are a lot of different ways to manage your venv.

A beginner will run the script, have it fail because of missing packages and read documentation.
A more experimented user will read the script to check what it does and probably use something like virtualenvwraper.

In the end, a simple line of documentation with a couple of links would do the job, less code to maintain :)

FlavioAlexander:
@lsenta Agree with you on those points, let's just keep them around until the new servers are stable, then we can modify fab scripts to do the same job,  probably within fewer lines of code. If you revert that, and fix the merge conflicts would be great. Then, I will need to add some misleading info based on the latest changes to have something up to date

lsenta:
@FlavioAlexander fixed & added back the setup scripts.
I also added the config and documentation for the new SBS_SHARED_ACCESS..., please confirm that covers everything that changed with the config.

Issue #1607, opened 2016-10-23
Several and different errors when submitting


gautier-marti:
I try to create a competition, it seemed to work fine, but when I try to submit an answer.zip containing a file answer.txt to test it, the submission fails: it displays several and different error messages for the same file submitted successively.

For example, I get :

```
Traceback (most recent call last):
  File "/home/azureuser/codalab/codalab/codalabtools/compute/worker.py", line 348, in run
    env=os.environ
  File "/usr/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1327, in _execute_child
    raise child_exception
OSError: [Errno 13] Permission denied
```

or 

```
Traceback (most recent call last):
  File "/home/azureuser/codalab/codalab/codalabtools/compute/worker.py", line 348, in run
    env=os.environ
  File "/usr/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1327, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

or
```
Traceback (most recent call last):
  File "/home/azureuser/codalab/codalab/codalabtools/compute/worker.py", line 255, in run
    shutil.rmtree(file_path)
  File "/usr/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/usr/lib/python2.7/shutil.py", line 256, in rmtree
    onerror(os.rmdir, path, sys.exc_info())
  File "/usr/lib/python2.7/shutil.py", line 254, in rmtree
    os.rmdir(path)
OSError: [Errno 39] Directory not empty: '/codalabtemp/tmpo3gREj/run'
```

To be sure that it was not my scoring program which was buggy, I replaced it with an "example" scoring program, and get same error messages. Here is the scoring program:

```
#!/usr/bin/env python
import sys
import os
import os.path

input_dir = sys.argv[1]
output_dir = sys.argv[2]

submit_dir = os.path.join(input_dir, 'res')
truth_dir = os.path.join(input_dir, 'ref')

if not os.path.isdir(submit_dir):
    print "%s doesn't exist" % submit_dir

if os.path.isdir(submit_dir) and os.path.isdir(truth_dir):
    if not os.path.exists(output_dir):
        os.makedirs(output_dir)

[...]

PR #1611, opened 2016-11-04
Solving partially #1607 (OSError: [Errno 39] Directory not empty)

julsal:
According to the post below, the "Directory not empty" error occurs because shutil.rmtree can't delete files set as read-only. The suggested solution was added in order to solve this problem.

IMPORTANT: Unfortunately, I don't have a Windows environment to fully test.

http://stackoverflow.com/questions/4829043/how-to-remove-read-only-attrib-directory-with-python-in-windows

### DIFF
--- a/codalab/codalabtools/compute/worker.py
+++ b/codalab/codalabtools/compute/worker.py
@@ -15,6 +15,7 @@ import signal
 import math
 import shutil
 import socket
+import stat
 import subprocess
 import sys
 import tempfile
@@ -34,6 +35,14 @@ from codalabtools import BaseWorker, BaseConfig
 from codalabtools.azure_extensions import AzureServiceBusQueue
 
 logger = logging.getLogger('codalabtools')
+
+
+def on_rm_error( func, path, exc_info):
+    # path contains the path of the file that couldn't be removed
+    # it assumes that it's read-only and unlink it.
+    os.chmod(path, stat.S_IWRITE)
+    os.unlink(path)
+
 
 class WorkerConfig(BaseConfig):
     """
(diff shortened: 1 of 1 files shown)


FlavioAlexander:
@julsal Even though this PR will solve the problem you mentioned above. I am afraid that the root of our problem is different. I have done some digging into this matter myself and found that  `.fuse_hidden files` are created, reason why `shutil.rmtree` won't be able to remove them. I have used `lsof` to find which process is using this file without any luck. The only thing that seems to work is `stopping/starting` the python process.

julsal:
I understand. In my case, 100% of the submissions fail with this error. My python code has more than one files. Do you think there is any relation to that? If you have difficulties to reproduce the problem, I can share my case with you.

FlavioAlexander:
@julsal I have never seen this error locally. Could you walk me through your set up? What OS are you  running on? If your machine is Windows, then your solution might work. Sorry for the late response :)

julsal:
@FlavioAlexander Just to make it clear, I didn't run the server on my side. I've just faced the error using it. If it still matters, I'm submitting from Ubuntu. If it helps, my suggestion was to share with you my competition metadata to allow you to reproduce the error.

FlavioAlexander:
@julsal Thanks for the clarification. What do you mean with your python code has more than one file? That shouldn't be a problem honestly. Is your competition being hosted on codalab server? Where is your competition hosted?

julsal:
@FlavioAlexander It means that my python script calls other python scripts present in the same directory. Yes, my competition is hosted on codalab server. It is this one: https://competitions.codalab.org/competitions/15363?secret_key=cfb32c59-5791-454f-be9f-b7c0013fed58

I just check the results and found out that the last submission succeeded.

FlavioAlexander:
@julsal Having multiple script shouldn't matter for this specific error. Please, try making more submissions and let me know that results. Recently, I set up new servers and haven't seen this error.

julsal:
@FlavioAlexander after some tests, it seems solved. Thank you very much for your help.

FlavioAlexander:
@julsal I am glad it is working, let me know if you have any more questions

Issue #1624, opened 2016-12-06
Incorrect "manage" link upon uploading a competition bundle


logological:
After successfully uploading a new competition bundle, the web interface prints the following message:

> Congratulations! Your new competition is ready to <a href='https://competitions.codalab.org/competitions/SOME_ID'>view</a>. You can also manage it from <a href='https://competitions.codalab.org/my/#manage'>your CodaLab dashboard.</a>

The problem is that the second hyperlink goes to `/my/#manage`, whereas there is no such anchor on that page.  If the user clicks on it, he or she will be directed to the main "My Competitions" page, which is still, rather inconveniently, one click away from the tab for managing the user's own competitions.  The actual anchor is `#my_managing`.

FlavioAlexander:
Fixed

PR #1625, opened 2016-12-06
Update invalid "manage" link anchors.  Fixes Issue #1624.

logological:
This pull request hasn't been tested but seems trivial enough.


FlavioAlexander:
@logological thanks for the PR, those changes should be enough to fix the links

PR #1675, opened 2017-02-05
Fixes #1674: Better link target for + button

### LINKED ISSUE #1674: Pressing "+" on the Submissions page jumps to top of page
In each row of the table on the Submissions page is a "+" button that, when pressed, reveals some download links for the submission.  Unfortunately, pressing it also causes the browser to jump back to the top of the page.  Along with Issue #1632, this makes it extremely inconvenient to download all the submission data.

The reason for the behaviour is that the target of the "+" link is `#`.  It would be better if the target were instead `#xxxxx`, where *xxxxx* is the value of the `id` attribute of the `tr` element containing the link.

logological:
Untested (but fairly obvious) fix for Issue #1674.  Made the link target for the + button the enclosing `tr` element.

### DIFF
--- a/codalab/apps/web/templates/web/my/submissions.html
+++ b/codalab/apps/web/templates/web/my/submissions.html
@@ -86,7 +86,7 @@
                                 <td>{{submission|get_item:column.name}}</td>
                             {% endif %}
                         {% endfor %}
-                        <td align="center"><a href="#{{submission.id}}" onclick="Competition.showOrHideSubmissionDetails(this)"><span class="glyphicon glyphicon-plus"></span></a></td>
+                        <td align="center"><a href="#{{ submission.id }}" onclick="Competition.showOrHideSubmissionDetails(this)"><span class="glyphicon glyphicon-plus"></span></a></td>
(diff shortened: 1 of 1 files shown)


codalab/apps/web/templates/web/my/submissions.html
```diff
@@ -86,7 +86,7 @@
                                 <td>{{submission|get_item:column.name}}</td>
                             {% endif %}
                         {% endfor %}
-                        <td align="center"><a href="#" onclick="Competition.showOrHideSubmissionDetails(this)"><span class="glyphicon glyphicon-plus"></span></a></td>
+                        <td align="center"><a href="#{{submission.id}}" onclick="Competition.showOrHideSubmissionDetails(this)"><span class="glyphicon glyphicon-plus"></span></a></td>
```
ckcollab: One really minor request, could you add spaces, like `{{ submission.id }}`

logological: @ckcollab: OK, done.

ckcollab:
Thanks!

PR #1677, opened 2017-02-07
CSV download for admins/creator

### LINKED ISSUE #1676: "Download CSV" button returns blank page instead of CSV
When a phase's "Leaderboard mode" is set to "Hide results", following the "Download CSV" link on the "Submissions" page returns a blank page instead of a CSV (even when there are several "Finished" submissions to that phase).  This behaviour is confusing.  I don't think it makes sense for "Hide results" to prevent the task organizer from downloading the results CSV.  But even if it does, the proper way of doing this would be to remove the "Download CSV" link altogether (or better yet, replace it with a message that results are not downloadable because "Hide results" has been set).

FlavioAlexander:
fixes #1676 

## Short summary 

- Since Submissions view is only accessible by admin/creators, they should be able to Download CSV files. This is a behavior we already have in other places.


ckcollab:
Thanks for this fix @FlavioAlexander!

PR #1687, opened 2017-02-17
Competition organizer feature

lotusgraham:
https://www.pivotaltracker.com/story/show/140153171
Adding a field to edit competitions page to begin setting up private compute worker.. this feature will allow organizers to specify a `vhost` for their compute worker, so only their compute worker will process these submissions.
screen shot 2017-02-17 at 4 30 08 pm

# Todo

- [x] Add some kind of permissions check when receiving submission updates... maybe we can use vhost as key?


ckcollab:
@madclam what do you think of this? Using different vhosts with secret keys, we can assign competitions to certain workers.

ckcollab:
Tweaking settings around trying to figure out how to get permissions to work... had to make a new queue type to receive submission updates from untrusted compute workers..

image

^ for reference later

PR #1694, opened 2017-03-08
Makes watch a dev dependency

lotusgraham:
This changes package.json so that watch is a dev dependency, since that feature's not necessary for building in production.   still get to watch by running `npm run watch-css`


ckcollab:
👍

PR #1704, opened 2017-03-20
Docker nice error pages

ckcollab:
- [x] Displays "maintenance" page for 503 errors
- [x] Generic error page for rest of errors
- [x] ~~Make nginx use include?~~ scope creep


lotusgraham:
right off the bat, I get this Warning about MYSQL variable. What works for me is that I can checkout another working branch, run `docker-compose up -d` and then checkout this branch, and then I can the site on localhost.
image

ckcollab:
So that looks like a problem with your `.env` ??

ckcollab:
Did you do `docker-compose stop` or `docker-compose stop django`...? You were supposed to stop the django server, so nginx will display an error page because it cannot reach django.

PR #1707, opened 2017-03-21
Xbaro/teams changes

ckcollab:


xbaro:
Your fix looks nice. I had no problems with current deployment, so I guess it works.

PR #1708, opened 2017-03-21
Docker db

xbaro:


codalab/codalab/settings/base.py
```diff
@@ -549,6 +590,8 @@ def post_setup(cls):
         if not hasattr(cls,'SERVER_NAME'):
             raise AttributeError("SERVER_NAME environment variable required")
 
+        if cls.SERVER_NAME not in cls.ALLOWED_HOSTS:
+            cls.ALLOWED_HOSTS.append(cls.SERVER_NAME)
```
ckcollab: what does this do? I just ran into a problem on Windows and we needed to adjust the `ALLOWED_HOSTS` -- I assume this automatically sets it nicely based on current host?

PR #1712, opened 2017-03-29
Env Sample Update

lotusgraham:


.env_sample
```diff
@@ -15,8 +15,9 @@ AWS_SECRET_ACCESS_KEY=
 AWS_STORAGE_BUCKET_NAME=public
 AWS_STORAGE_PRIVATE_BUCKET_NAME=private
 AWS_S3_CALLING_FORMAT=boto.s3.connection.OrdinaryCallingFormat
-AWS_S3_HOST=s3-us-west-2.amazonaws.com
+AWS_S3_HOST=s3.amazonaws.com
 AWS_QUERYSTRING_AUTH=False
+S3DIRECT_REGION=us-west-2
```
ckcollab: can we just set this to `S3DIRECT_REGION=s3`? Was that the one that worked?

lotusgraham: na, that didn't work. we only used 'S3' for 'AWS_S3_HOST`

ckcollab: kk, hmm, I'll leave it up to you whether you want to do a little bonus assignment: can you parse out the `"us-west-2"` or `"us-east-1"` parts of the AWS settings so users only have to set that once? If not, just a little comment is probably good enough.

ckcollab: Nevermind, I guess it doesn't make sense if it's set to `AWS_S3_HOST=s3.amazonaws.com` because that won't work =\

ckcollab:
👍

PR #1718, opened 2017-04-05
Added cookie advisor for compliance with EU

xbaro:


ckcollab:
Tested locally, works well! 👍

PR #1722, opened 2017-04-12
Adds user entries and date to leaderboard

lotusgraham:
pasted image at 2017_04_10 10_25 pm

### DIFF
--- a/codalab/apps/web/views.py
+++ b/codalab/apps/web/views.py
@@ -679,6 +679,16 @@ class CompetitionResultsPage(TemplateView):
             context['is_owner'] = is_owner
             context['phase'] = phase
             context['groups'] = phase.scores()
+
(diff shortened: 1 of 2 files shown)


ckcollab:
Could you make the "Results" table header stretch the full width?

codalab/apps/web/views.py
```diff
@@ -679,6 +679,16 @@ def get_context_data(self, **kwargs):
             context['is_owner'] = is_owner
             context['phase'] = phase
             context['groups'] = phase.scores()
+
+            for group in context['groups']:
```
ckcollab: Could you add a ton of submissions and see if this causes a huge amount of lag? Or, at least make a note of this and when we push to testing, make sure a competition with lots of  submissions isn't lagging?

PR #1730, opened 2017-04-17
Add staging server instructions to readme

jared-cain:


ckcollab:
Can we move this to docs and take out hard coded IP? Maybe just say "reference the slack pinned message for the .pem file and IP"?

Thanks :D

ckcollab:
Will re-open if @jared-cain wants to

Issue #1747, opened 2017-05-10
Turn on/off special style [see.4c]


dragonlocke:
Admin editor can turn on/off special style. (special admin page to set up single competition mode)

dragonlocke:
![image](https://cloud.githubusercontent.com/assets/25310715/26023996/4518910a-377d-11e7-8b94-edfc6cc72519.png)
Is this what is implied by special style? If not is it in the competition edit page? Also is it on 

https://competitions-test.codalab.org/ or http://docker-server.lri.fr/ ?

dragonlocke:
Found in http://docker-server.lri.fr/

Page currently not functional, submitting a style change results in error page.

ckcollab:
@dragonlocke I believe this is fixed now?

dragonlocke:
-uploading image works
-uploading new image over previous image works

However:
-can't upload competitino still, therefore cannot test single competition mode
-clear image tickbox next to the uploaded image if checked and saved results in error page

![image](https://cloud.githubusercontent.com/assets/25310715/26080039/aed527d6-3979-11e7-84bf-81bb958cbee1.png)

dragonlocke:
case closed

PR #1767, opened 2017-05-11
Single competition mode editor/customization

ckcollab:
image

Solves #1747

# Todo

- [x] Middleware for handling env var OR `Configuration` competition pk
- [x] Change logo based on global `Configuration` object

### DIFF
--- a/codalab/apps/customizer/admin.py
+++ b/codalab/apps/customizer/admin.py
@@ -1,0 +1,5 @@
+from django.contrib import admin
+import models
+
+
+admin.site.register(models.Configuration)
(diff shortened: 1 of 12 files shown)


ckcollab:
@dragonlocke I'm merging + deploying this now, you need to be an admin to test it.. let me know on slack what your name is and such on the docker staging server and I'll set you as an admin so you can tweak this.

To test this, when you're an admin, you'll see **Customize Codalab** in your user dropdown.

image

Then you'll see the form 

image


And you can tweak from there

dragonlocke:
@ckcollab username: wlutz

PR #1803, opened 2017-05-21
prevent textarea resizing

dominiquemb:


dominiquemb:
Whoops, need to make a change to this

PR #1804, opened 2017-05-21
Delete submission button alignment

dominiquemb:


dominiquemb:
This branch was based off of master which is causing some problems with docker-compose. Need to create a new branch based off develop

ckcollab:
No big deal, let's delete old branches though so they don't muddy up things

PR #1805, opened 2017-05-22
100% width

dominiquemb:


ckcollab:
@dominiquemb @dragonlocke can we QA this one first?

PR #1807, opened 2017-05-22
Delete submission button alignment

dominiquemb:


codalab/apps/web/templates/web/my/submission_delete.html
```diff
@@ -10,7 +10,7 @@
 <form action="" method="post">
     {% csrf_token %}
     <p>Are you sure you want to delete this submission?</p>
-    <input type="submit" value="Confirm" class="button alert confirm-button btn btn-primary"/>
```
ckcollab: Good catch... what the heck!

PR #1826, opened 2017-06-02
Doc updates 2

dominiquemb:


ckcollab:
image

Having these as menu items is kinda weird -- is that why you were using `<h2>` and such?

Issue #1810, opened 2017-05-24
Set custom docker image to run submissions


ckcollab:

ckcollab:
Actually Tyler mostly QAd this so I think we can mark it closed.

PR #1837, opened 2017-06-14
Use custom docker images to run submissions

ckcollab:
- [x] ~~Should be a flag organizers can enable or disable ???~~ Holding off on this
- [x] ~~Create Codalab docker hub organization, talk to @xbaro~~ No response, don't want to bug him too much more.. has enough things on his plate :)
- [x] Build custom docker image representing old style compute workers (INSTALL **JAVA**)
- [x] Make a docker image input field that is shown/hidden when appropriate
- [x] Actually use the docker image when processing submission
- Set a default "docker image" for yourself? Or remember it between pages??
    - [x] use `localStorage` to generate a key based on the competition ID (so we can store docker image names PER competition)
    - [x] Load the input field with value from `localStorage` key
    - [x] When submit button is clicked, `localStorage` is saved
- [x] How should I clean up old containers....?
    - Looks like we can do this with `docker rm $(docker ps -qa --no-trunc --filter "status=exited")` but... when should we run this? Do we need to setup a cron job? Can we add this to a docker container somewhere.......
    - Also, `docker container prune -f` https://docs.docker.com/engine/reference/commandline/container_prune/ could be useful?
    - NOTE there is also `system prune` which deletes more than `container prune`

Completes #1810


ckcollab:
NOTE: Tyler working on `localStorage` and I'm trying to make the legacy docker image

ckcollab:
I think system prune will be pretty useful...!

image

codalab/apps/web/templates/web/competitions/_submit_results_page.html
```diff
@@ -140,6 +144,24 @@
 {% endif %}
 
 <script>
+
```
ckcollab: Can we add a comment here explaining what the following code does?

ckcollab: Maybe make it sections, like...

```javascript
// ====================================
// Docker image input local storage

..

// ====================================
// s3 file upload

..

```

codalab/apps/web/templates/web/competitions/_submit_results_page.html
```diff
@@ -140,6 +144,24 @@
...
+    var docker_image_key = window.location.href + '_docker_image'
+
+    $(document).ready(function () {
+        $(window).unload(saveSettings);
+        loadSettings();
+    });
+
+    function loadSettings() {
```
ckcollab: annoying nitpick: I've been trying to do all new variables/functions in JS lowercase_separated_by_underscores.

(another annoying nitpick:  no semicolons  :))

PR #1840, opened 2017-06-15
1777 mysql to postgres

Tthomas63:
Sets up branches to use postgres instead of mysql.

Note: The following must be set in .env

```
DB_HOST=postgres
DB_PORT=5432
DB_NAME=codalab_website
DB_USER=root
DB_PASSWORD=password -- Change to your setup.
DB_DATA_PATH=./var/data/postgres

DB_ENGINE = postgresql -- This variable is already present, just change it to postgresql.
```
Also note in order to dump old competitions/users from mysql before switching to postgres ( this branch or pulling the latest once this gets merged ) use the following commands.:
`python manage.py dumdata web authenz --exclude=contenttypes --natural > db_test_dump.json` 
to dump the data for the web and authenz apps.

Once switched over to using postgres use the following on a fresh db ( See below for explanation on creating/destroying db )
```
python manage.py syncdb --migrate --no-initial-data --noinput
python manage.py sqlflush
python manage.py loaddata db_test_dump.json
```
Note: db_test_dump.json can be named anything for dumping/loading

For making sure your database is fresh/creating/destroying:
```
docker exec -it postgres bash
dropdb codalab_website
createdb codalab_website
exit -- To get out of postgres docker container
```

The above process was tested using the staging server mysql database, and was able to be loaded into my local codalab instance.

### DIFF
--- a/codalab/codalab/settings/base.py
+++ b/codalab/codalab/settings/base.py
@@ -522,11 +522,11 @@ class Base(Settings):
                 DATABASES = {
                     'default': {
                         'ENGINE': 'django.db.backends.postgresql_psycopg2',
-				        'NAME': os.environ.get('DB_PG_NAME', 'postgres'),
-				        'USER': os.environ.get('DB_PG_USER', 'postgres'),
-				        'PASSWORD': os.environ.get('DB_PG_PASSWORD', ''),
-				        'HOST': os.environ.get('DB_PG_HOST', 'postgres'),
-				        'PORT': os.environ.get('DB_PG_PORT', '5432'),
+                        'NAME': os.environ.get('DB_NAME', 'codalab_website'),
+                        'USER': os.environ.get('DB_USER', 'postgres'),
+                        'PASSWORD': os.environ.get('DB_PASSWORD', ''),
+                        'HOST': os.environ.get('DB_HOST', 'postgres'),
+                        'PORT': os.environ.get('DB_PORT', '5432'),
                     }
                 }
             elif db_engine == 'sqlite3':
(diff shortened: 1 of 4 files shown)


codalab/codalab/settings/base.py
```diff
@@ -568,6 +568,15 @@ class Base(Settings):
                     }
                 }
 
+    DATABASES['postgres'] = {
```
ckcollab: we want to pull this, I think?

Tthomas63: Ah yeah sorry thought this already got pulled out.

docker-compose.yml
```diff
@@ -38,6 +38,20 @@ services:
...
 
+  postgres:
+    image: postgres:9.6.3
+    environment:
+      - POSTGRES_DB=${DB_NAME}
+      - POSTGRES_USER=${DB_USER}
+      - POSTGRES_PASSWORD=${DB_PASSWORD}
+    volumes:
+      - ./docker:/app/docker
+      - ./docker/psql:/etc/psql/conf.d
```
ckcollab: Is this necessary any more? we don't have a `psql/conf.d`? Can maybe leave it here in case we do want one later? Or put a blank placeholder file there so it's less confusing?

Tthomas63: Hmm not sure. Its probably not necessary for now until we do something with it in the future.

ckcollab: Cool, I'll leave it up to you to decide (pull it, or add blank file)

ckcollab:
great stuff thank you tyler! :shipit:

PR #1843, opened 2017-06-20
Fixes download all submissions in comp submissions.

Tthomas63:
Fixes the download all button in the submissions management page of a competition. 

The error arose from trying to treat S3 files as azure storage files.


ckcollab:
:shipit:

Issue #1851, opened 2017-06-26
Only owners of the submission  or competition creator can delete submissions


xbaro:
In the administration view of the submissions, if you are an administrator of the competition, but not the creator, you cannot delete submissions. 

To fix it, is necessary to check for administrators on the SubmissionDelete view of the web application:

From:
```python
    class SubmissionDelete(LoginRequiredMixin, DeleteView):
        model = models.CompetitionSubmission
        template_name = "web/my/submission_delete.html"

        def get_object(self, queryset=None):
            obj = super(SubmissionDelete, self).get_object(queryset)
   
            self.success_url = reverse("competitions:view", kwargs={"pk": obj.phase.competition.pk})

            if obj.participant.user != self.request.user and obj.phase.competition.creator != self.request.user:
                raise Http404()

            return obj
```

To:

```python
    class SubmissionDelete(LoginRequiredMixin, DeleteView):
        model = models.CompetitionSubmission
        template_name = "web/my/submission_delete.html"

        def get_object(self, queryset=None):
            obj = super(SubmissionDelete, self).get_object(queryset)

            is_admin = self.request.user in obj.phase.competition.admins.all()
            self.success_url = reverse("competitions:view", kwargs={"pk": obj.phase.competition.pk})

            if obj.participant.user != self.request.user and obj.phase.competition.creator != self.request.user and not is_admin:
                raise Http404()

            return obj
```

Tthomas63:
Implemented in PR #1854

PR #1854, opened 2017-06-27
FIX Only owners of the submission or competition creator can delete submissions

Tthomas63:
This implements Xbaro's fix for #1851

### DIFF
--- a/codalab/apps/web/views.py
+++ b/codalab/apps/web/views.py
@@ -1403,11 +1403,11 @@ class SubmissionDelete(LoginRequiredMixin, DeleteView):
         obj = super(SubmissionDelete, self).get_object(queryset)
 
         is_admin = self.request.user in obj.phase.competition.admins.all()
-        self.success_url = reverse("competitions:view", kwargs={"pk": obj.phase.competition.pk})
-
+        # Check if user is not equal to the current user, competition creator, or admin
         if obj.participant.user != self.request.user and obj.phase.competition.creator != self.request.user and not is_admin:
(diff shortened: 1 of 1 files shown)


codalab/apps/web/views.py
```diff
@@ -1402,14 +1402,14 @@ class SubmissionDelete(LoginRequiredMixin, DeleteView):
     def get_object(self, queryset=None):
         obj = super(SubmissionDelete, self).get_object(queryset)
 
+        is_admin = self.request.user in obj.phase.competition.admins.all()
         self.success_url = reverse("competitions:view", kwargs={"pk": obj.phase.competition.pk})
 
-        if obj.participant.user != self.request.user and obj.phase.competition.creator != self.request.user:
+        if obj.participant.user != self.request.user and obj.phase.competition.creator != self.request.user and not is_admin:
```
ckcollab: Can we add a comment for this logic?

codalab/apps/web/views.py
```diff
@@ -1402,14 +1402,14 @@ class SubmissionDelete(LoginRequiredMixin, DeleteView):
     def get_object(self, queryset=None):
         obj = super(SubmissionDelete, self).get_object(queryset)
 
+        is_admin = self.request.user in obj.phase.competition.admins.all()
         self.success_url = reverse("competitions:view", kwargs={"pk": obj.phase.competition.pk})
```
ckcollab: Can we move this after the if? So the `is_admin` stuff is grouped, then `self.success_url` is set afterwards?

PR #1869, opened 2017-07-11
Docker automatic pruning

Tthomas63:
- [X] Creates a new variable in settings `DOCKER_MAX_SIZE_GB` ( Default is 10.0 GB )
- [X] Checks docker images size, if greater than the above settings, forces a prune.
- [X] Branched off of develop instead of the branch for ensuring legacy competition stability with a submission script.

image


Tthomas63:
@ckcollab This is ready for review

codalab/apps/web/templates/web/competitions/_submit_results_page.html
```diff
@@ -151,7 +151,7 @@
            var maxLength = 128;
            var text = $(this).val();
            var textLength = text.length;
```
ckcollab: while we're here, can we change this variable from CamelCase to snake_case?

codalab/codalabtools/compute/worker.py
```diff
@@ -55,6 +57,16 @@ def _find_only_folder_with_metadata(path):
                 return folder
 
 
+def docker_prune():
+    """Runs a prune on docker if our images take up more than what's defined in settings."""
+    docker_image_size = os.popen("docker system df | awk -v x=4 'FNR == 2 {print $x}'").read().strip()
```
ckcollab: can get rid of `docker_` prefixing all of these variable names

ckcollab: Also, let's pair today and write tests for this

PR #1893, opened 2017-08-04
Added drills to docs

Tthomas63:
Adds the test drills from Isabelle to `readthedocs` and includes more drills for some newer features.


ckcollab:
Thanks a ton, modified them a bit, pushing!

PR #1907, opened 2017-08-22
Migration to update page ranks

Tthomas63:
Adds a migration to check page's ranks, and update them if necessary to reflect the inclusion of starting_kits/files.


ckcollab:
Could you test this on demo data before merging? You could check this branch out on the docker server?

Otherwise :shipit:

Tthomas63:
Can do. Already tested locally, and I believe I have the same data. Will check the results more thoroughly and test on docker staging as well.

Tthomas63:
This is working docker-staging. Added a few more tweaks such as updating the starting kit page label if it was not set to Files.

PR #1908, opened 2017-08-25
Public Data OrganizerDatasets

Tthomas63:
Adds:
- [x] `public_data` tag to competition yaml. You may use a filename in the bundle, or a URI key.
- [x] `public data` organizer dataset type
- [x] `public_data` fields on phases
- [x] Appropriate helper functions to get size, and download url. These should be re-written to generically handle files passed.

image

*Note: Screenshot shows type as reference data because I used an already uploaded Dataset as an example*


codalab/apps/web/models.py
```diff
@@ -1801,6 +1830,35 @@ def dict_constructor(loader, node):
                         assert False, "OrganizerDataSet (%s) could not be found" % phase_spec["starting_kit"]
                         # End unpack starting kit
 
+            # Begin unpack public data
+            if hasattr(phase, 'public_data') and phase.public_data:
+                if phase_spec["public_data"].endswith(".zip"):
```
ckcollab: this doesn't handle multiple `public_datas` I think?

Tthomas63: No, is that something we want? It might be a little redundant with one public data field per Phase, and a starting kit field per phase.

ckcollab: Yep yep, if we're replacing that old `data` stuff, that handled multiple External Files or whatever right? I sent you some DMs about a plan to add a new model to point many `PublicDatas` at 1 competition

Tthomas63: Isabelle decided against this, since we can group together organizer datasets.

PR #1949, opened 2017-09-19
Single comepetition view for just those competitions

ckcollab:


ckcollab:
We could use some more input from @madclam here. We have changed the template but I don't think this is what you want -- it just shows/hides a couple small things. It's almost imperceptible.

image


There's no "My Competitions" link, for example

PR #2037, opened 2017-11-06
Comment out HTMl to select submission docker image

Tthomas63:
- [x] Comments out HTML/JS for participants selecting a custom docker image. ( Organizers may still set a docker image or use the Codalab default. )


ckcollab:
thanks bud

PR #2038, opened 2017-11-06
Phase migrations: Submission docker images.

Tthomas63:
- [ ] Submission docker_images should carry over now with migration. Simply needed to tell the new submission on the phase to use docker image.


ckcollab:
Thank you very much!

PR #2039, opened 2017-11-06
Detailed Results Empty/Metadata file fix

Tthomas63:
- [ ] Detailed results are now displayed properly in the `iframe`. Does not download, but displays.
   - Use existing view instead of sass URL


Tthomas63:
Pushed to `http://codalab15-azure.eastus.cloudapp.azure.com/`. Working, and displays as follows for the following competition/submission: 

- Doesn't download
- Displays in iFrame

http://codalab15-azure.eastus.cloudapp.azure.com/competitions/49#results

screen shot 2017-11-07 at 1 41 25 pm
screen shot 2017-11-07 at 1 41 16 pm

ckcollab:
Code looks OK to merge to develop, so we don't accidentally not have this on a server when Isabelle goes to test it.

PR #2040, opened 2017-11-07
Clean Script tags from HTML

Tthomas63:
Adds a utility function that removes script tags and anything in-between on competition pages.

This gets called when
- We edit the competition from the form
- We upload raw HTML
- Attempt to save a Page object.


Tthomas63:
Tests added!

ckcollab:
So putting whitespace after `<` breaks this.

We should use this regex, it sounds like:

```/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi```

That's what jQuery uses to filter out script tags

Issue #2049, opened 2017-11-14
Set a competition-wide docker, instead of fine-grained dockers


ckcollab:
Instead of having a docker image for:

 - participant code
 - scoring program
 - ingestion program

Just have 1 default docker image for the competition.

Tthomas63:
See pull request #2050

Tthomas63:
Currently on http://codalab15-azure.eastus.cloudapp.azure.com/ for QA/Review

madclam:
I ran the following tests:
1) In competition http://codalab15-azure.eastus.cloudapp.azure.com/competitions/87
- changed with editor Competition docker image: madclam/codalab:anaconda.4.4.0
- re-ran submission 729 in the "Submissions" admin panel (this gave submission 738).
As can be seen, submission 738 did NOT get run into docker madclam/codalab:anaconda.4.4.0
![image](https://user-images.githubusercontent.com/7594063/32864239-34474b5e-ca13-11e7-83b4-ac02611c60c0.png)

2) In the same competition, I re-uploaded the same submission directly from "Submit/View results" WITH madclam/codalab:anaconda.4.4.0 as participant docker [NOTE: THIS FEATURE SHOULD BE DISABLED!] This ran fine in the docker specified as:
![image](https://user-images.githubusercontent.com/7594063/32864321-94d465e2-ca13-11e7-94fb-440038c1b6bf.png)

3) I redid the same thing WITHOUT specifying a participant docker and got:
![image](https://user-images.githubusercontent.com/7594063/32864431-f426997a-ca13-11e7-90ed-7b369c7fcfb3.png)

4) I changed again the organizer docker with the editor to:
- Competition docker image: lisesun/codalabautoml2016:3.0
- I made again the sample submission via "Submit/View results"
it failed with 
Unable to find image 'lisesun/codalabautoml2016:3.0' locally
docker: Error response from daemon: pull access denied for lisesun/codalabautoml2016, repository does not exist or may require 'docker login'.
See 'docker run --help'.

5) Replaced Lisheng's docker by continuumio/anaconda:latest
- Competition docker image: continuumio/anaconda:latest
- I made again the sample submission via "Submit/View results"
This ran in the correct docker
![image](https://user-images.githubusercontent.com/7594063/32865540-6b06115c-ca18-11e7-905c-11374178eca9.png)

6) I ran the same thing, but this time:
[...]

PR #2050, opened 2017-11-14
Add competition docker image

Tthomas63:
Fixes issue #2049 

Adds a field to competition for docker images. All other docker fields are removed from the front-end, but are accessible on the backend.

Also modified re-running submissions/migrating submissions to use the submissions docker image.

If a submission has a docker image already associated, it uses that. If it does not, it uses the competition docker if set, or the default docker specified in settings. 

Parsing submissions has the following behavior:

screen shot 2017-11-14 at 12 59 32 pm
screen shot 2017-11-14 at 11 58 43 am

### DIFF
--- a/codalab/apps/web/forms.py
+++ b/codalab/apps/web/forms.py
@@ -45,6 +45,7 @@ class CompetitionForm(forms.ModelForm):
             'anonymous_leaderboard',
             'enable_teams',
             'require_team_approval',
+            'competition_docker_image',
         )
         widgets = {'description': TinyMCE(attrs={'rows' : 20, 'class' : 'competition-editor-description'},
                                           mce_attrs={"theme": "advanced", "cleanup_on_startup": True, "theme_advanced_toolbar_location": "top", "gecko_spellcheck": True})}
@@ -85,14 +86,10 @@ class CompetitionPhaseForm(forms.ModelForm):
             'scoring_program_organizer_dataset',
             'phase_never_ends',
             'force_best_submission_to_leaderboard',
-            'scoring_program_docker_image',
-            'default_docker_image',
-            'disable_custom_docker_image',
             'ingestion_program_organizer_dataset',
-            'ingestion_program_docker_image',
         )
         labels = {
-            'default_docker_image': "Default participant docker image"
+            # 'default_docker_image': "Default participant docker image"
         }
         widgets = {
             'leaderboard_management_mode' : forms.Select(
--- a/codalab/apps/web/forms.py
+++ b/codalab/apps/web/forms.py
@@ -88,9 +88,6 @@ class CompetitionPhaseForm(forms.ModelForm):
             'force_best_submission_to_leaderboard',
             'ingestion_program_organizer_dataset',
         )
-        labels = {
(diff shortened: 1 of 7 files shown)


codalab/apps/web/forms.py
```diff
@@ -85,15 +86,8 @@ class Meta:
             'scoring_program_organizer_dataset',
             'phase_never_ends',
             'force_best_submission_to_leaderboard',
-            'scoring_program_docker_image',
```
ckcollab: can we just comment these out or something instead?

codalab/apps/web/models.py
```diff
@@ -258,6 +258,8 @@ class Competition(models.Model):
     require_team_approval = models.BooleanField(default=True, verbose_name="Organizers need to approve the new teams")
     teams = models.ManyToManyField(Team, related_name='competition_teams', blank=True, null=True)
 
+    competition_docker_image = models.CharField(max_length=128, default='', blank=True)
```
ckcollab: `default=settings.DEFAULT_WORKER_IMAGE` or what-have-you

codalab/apps/web/tasks.py
```diff
@@ -225,23 +225,31 @@ def _prepare_compute_worker_run(job_id, submission, is_prediction):
...
+
+    if submission.docker_image and submission.docker_image != "":
+        docker_image = submission.docker_image
+    else:
+        docker_image = submission.phase.competition.competition_docker_image or settings.DOCKER_DEFAULT_WORKER_IMAGE
+        submission.docker_image = docker_image
+        submission.save()
+
+    logger.info("@@@ Docker image set to: {} @@@".format(docker_image))
+    print("@@@ Docker image set to: {} @@@".format(docker_image))
```
ckcollab: remove this print or whichever isn't useful

codalab/apps/web/tasks.py
```diff
@@ -737,6 +740,7 @@ def re_run_all_submissions_in_phase(phase_pk):
         new_submission = CompetitionSubmission(
             participant=submission.participant,
             phase=submission.phase,
+            docker_image=submission.docker_image,
```
ckcollab: ty good catches on this

codalab/apps/web/models.py
Tthomas63: This doesn't really make sense though in this case though I think? We check if the organizer has actually set something here rather than being blank. If we set it to the default we would always be comparing it to the setting-default to see if it was different later on in the code?

codalab/apps/web/forms.py
Tthomas63: This is now the case, I just believe git is not displaying the commented out code.

Tthomas63:
@ckcollab Some changes pushed

codalab/apps/web/forms.py
ckcollab: ty

codalab/apps/web/models.py
ckcollab: that makes sense

Tthomas63:
Currently on http://codalab15-azure.eastus.cloudapp.azure.com/ for QA/Review

PR #2101, opened 2017-11-29
Force scores to always return list format

Tthomas63:
Addresses an issue with the leaderboard, where when editing an active competition, or in some other unknown cause, the competition fails to display overall, or the results page fails to display, or the submissions page fails to display. 

The fix forces the `scores` method on phase, to always return in the same format. 

Tested with broken competitions locally, and is working. Unaffected competitions should remain unchanged.


ckcollab:
Thanks for digging into this mess @Tthomas63 !

Issue #2147, opened 2017-12-21
Docker image name stripping characters it shouldnt


ckcollab:

Tthomas63:
See #2166

PR #2166, opened 2018-01-16
Fix Docker Image Clean Stripping Hyphens

Tthomas63:
Fixes the exclusion rule for hyphens in docker image names.(#2147)


ckcollab:
:shipit:

PR #2169, opened 2018-01-18
Hide chart if phase is blind

Tthomas63:
Adds a conditional to hide the graph chart if the phase is blind.

### DIFF
--- a/codalab/apps/web/templates/web/competitions/view.html
+++ b/codalab/apps/web/templates/web/competitions/view.html
@@ -276,8 +276,12 @@
                 </div>
             {% endif %}
         </div>
-        <div class="col-md-8">
-            {% include "web/competitions/_chart.html" %}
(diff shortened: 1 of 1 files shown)


codalab/apps/web/templates/web/competitions/view.html
```diff
@@ -276,8 +276,12 @@ <h3 class="panel-title"><strong>Competition Ends</strong></h3>
                 </div>
             {% endif %}
         </div>
-        <div class="col-md-8">
-            {% include "web/competitions/_chart.html" %}
+        <div align="center" class="col-md-8">
+            {% if active_phase.is_blind %}
+                <i>This phase is blind</i>
```
ckcollab: Do we even need a message here if the phase is blind?

Tthomas63: Nope I guess not

Issue #2233, opened 2018-03-22
submissions possible beyond the competition deadline


madclam:
in the competition
https://competitions.codalab.org/competitions/18633
the end deadline is passed
but it is still possible to make submissions...

ckcollab:
Continue to allow phases to start at the same time, but show a warning or disallow phases to start after competition ends or phases to start out of order

PR #2237, opened 2018-03-27
Add check to submissions if they're past competition end

Tthomas63:
Addresses issue #2233 

Checks if the submission is past the competition deadline (If it exists). If so throw an error.


ckcollab:
:shipit:

PR #2248, opened 2018-04-01
Add check for if phase_never_ends and add tests for submissions past deadlines

Tthomas63:
Modify the check on submissions for past `competition.end_date` to exempt phases that never end

### DIFF
--- a/codalab/apps/web/tests/test_submission_past_deadlines.py
+++ b/codalab/apps/web/tests/test_submission_past_deadlines.py
@@ -1,0 +1,85 @@
+import datetime
+
+import pytest
+from django.core.exceptions import PermissionDenied
+from django.core.urlresolvers import reverse
+from django.test import TestCase, Client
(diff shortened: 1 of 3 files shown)


codalab/apps/web/tests/test_submission_past_deadlines.py
```diff
@@ -0,0 +1,85 @@
...
+
+        with pytest.raises(PermissionDenied) as exc_info:
+            self.submission_1 = CompetitionSubmission.objects.create(
+                participant=self.participant_1,
+                phase=self.phase_1,
+                status=submission_finished,
+                submitted_at=datetime.datetime.now()
+            )
+
+        exception_raised = exc_info.value
```
ckcollab: is this necessary? What happens when `pytest.raises` doesn't get the permission denied raised? is `exc_info` even available outside of that with? shouldn't need to do any of this if it doesn't raise it? I don't think anyway

PR #2267, opened 2018-04-18
A-Teams

Tthomas63:
Features for A-Teams:
 - Competition `Allow Organizer Teams` option
 - Remove Teams tab/participation approval
 - Add Create Organizer Team Button/Form/View/Template
 - Add Create Organizer Team from CSV Button/Form/View/Template
 - Added logic to `Team` save to prevent users from having more than 1 `TeamMembership` Object per `Competition`.
 - Add validation/error handling to forms

### DIFF
--- a/codalab/apps/teams/forms.py
+++ b/codalab/apps/teams/forms.py
@@ -1,13 +1,18 @@
 from django import forms
+from django.core.exceptions import ValidationError
+from django.db.models import Q
 from django.forms.widgets import ClearableFileInput, Input, CheckboxInput
 from django.utils.html import escape, conditional_escape
 from django.utils.encoding import force_unicode
-from .models import Team, TeamMembership
 from tinymce.widgets import TinyMCE
-from django.utils.html import format_html
-from django.utils.encoding import force_text
+from django.utils import timezone
+from apps.web.models import Competition
+from apps.authenz.models import ClUser
+from apps.teams.models import TeamMembership, Team, TeamStatus, TeamMembershipStatus
 from django.utils.safestring import mark_safe
 import os
+import datetime
+
 
 class CustomImageField(ClearableFileInput):
 
@@ -36,6 +41,7 @@ class CustomImageField(ClearableFileInput):
                 substitutions['clear_template'] = clear_template % substitutions
 
         return mark_safe(template % substitutions)
+
 
 class TeamEditForm(forms.ModelForm):
     name = forms.CharField(max_length=64, required=False)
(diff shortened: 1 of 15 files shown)


codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+
+    text_members = forms.CharField(required=False, widget=forms.Textarea)
+
+    class Meta:
+        model = Team
+        fields = ('description', 'name')
+
+    def __init__(self, *args, **kwargs):
+        self.user_list = []
+        self.competition = Competition.objects.get(pk=kwargs.pop('competition_pk'))
```
ckcollab: Why do this here? instead of when cleaning/validating? seems like a weird place to do it?

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+        self.creator = ClUser.objects.get(pk=kwargs.pop('creator_pk'))
+        super(OrganizerTeamForm, self).__init__(*args, **kwargs)
+        # This check is to see if we have an existing object with a PK, or if we're making a new one
+        # If we have an existing we grab existing members for the form.
+        if hasattr(self, 'instance'):
+            if hasattr(self.instance, 'pk'):
+                if self.instance and self.instance.pk:
+                    team = Team.objects.get(pk=self.instance.pk)
+                    current_members = team.members.all()
```
ckcollab: If you want the names just do `team.members.all().values_list('username', flat=True)`

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+                        name_list.append(member.username)
+                        name_text_list = ",".join(name_list)
+                    self.initial['text_members'] = name_text_list
+
+    def clean_text_members(self):
+        if self.cleaned_data.get('text_members'):
+            team_members_text = self.cleaned_data.get('text_members')
+            team_members_list = team_members_text.replace(' ','').split(',')
+            # team_members_list = team_members_text.replace(' ','').split('\n')
+            self.user_list = ClUser.objects.filter(Q(username__in=team_members_list) | Q(email__in=team_members_list))
```
ckcollab: i don't think we should be doing it like this, jumping from function to function kinda?

ckcollab: Probably should just do all of this in the [clean()](https://docs.djangoproject.com/en/1.7/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other) method

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+                    raise ValidationError(('Invalid value for name. A team already exists with that name.'), code='invalid')
+        return self.cleaned_data.get('name')
+
+    def save(self, commit=True):
+        self.instance.creator = self.creator
+        self.instance.competition = self.competition
+        self.instance.status = TeamStatus.objects.get(codename=TeamStatus.APPROVED)
+
+        # We need to call save before adding members
+        self.instance.save()
```
ckcollab:
```python
if not self.instance.pk: 
    self.instance.save()```

Maybe instead here? probably not too important

File: codalab/apps/teams/forms.py [hunk 5]
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+
+        if hasattr(self, 'user_list'):
+            # On save, if we have existing objects, clear it. We only add names in the form.
+            if len(self.instance.members.all()) > 0:
```
ckcollab: I wouldn't repeat the query like this... do:

```python
members = self.instance.members.all()
```

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+                # Remove users no longer in the list
+                for user in self.instance.members.all():
+                    if user not in self.user_list:
+                        # Delete members not on the new list
+                        print("Deleting membership for user: {}".format(user))
```
ckcollab: remove the `print` or turn it into a `log` proper

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+                        TeamMembership.objects.get(user=user, team=self.instance).delete()
```
ckcollab: nice all on one line 👍

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+            for user in self.user_list:
+                if user not in self.instance.members.all():
+                    # There is not already an existing membership for this user.
+                    new_team_membership = TeamMembership.objects.create(
+                        user=user,
+                        team=self.instance,
+                        is_invitation=False,
+                        is_request=False,
+                        start_date=timezone.now(),
+                        end_date=timezone.now() + datetime.timedelta(days=9000),
```
ckcollab: weird, can we just have a `null` `end_date`?

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+                        message="Organizer Created",
+                        status=TeamMembershipStatus.objects.get(codename=TeamMembershipStatus.APPROVED),
+                        reason="Organizer Created Team"
+                    )
+                    print("Created new membership: {0} for user: {1} on team: {2}".format(new_team_membership, user, self.instance))
```
ckcollab: Same comment on all the prints here, should be logs ideally

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+    csv_file = forms.FileField(required=False)
+
+    def __init__(self, *args, **kwargs):
+        self.competition = Competition.objects.get(pk=kwargs.pop('competition_pk'))
+        self.creator = ClUser.objects.get(pk=kwargs.pop('creator_pk'))
+        self.teams_dict = {}
+        super(OrganizerTeamsCSVForm, self).__init__(*args, **kwargs)
+
+    def clean_csv_file(self):
+        teams_dict = {}
```
ckcollab: I don't think you need the `self.field` class attributes, just return the value a la https://docs.djangoproject.com/en/1.7/ref/forms/validation/#cleaning-a-specific-field-attribute

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+        if self.cleaned_data.get('csv_file'):
+            temp_csv = self.cleaned_data.get('csv_file')
+            csv_team_list = [line.rstrip() for line in temp_csv]
+            for index, value in enumerate(csv_team_list):
+                temp_line = value.split(',')
+                temp_line = [s.strip() for s in temp_line]
+                teams_dict[temp_line.pop(0)] = temp_line
```
ckcollab: `pop(0)`?

Why not make it like...

```python
teams =  {}
for line in file:
    team_name, emails = line.split(",")
    teams[team_name.strip()] = [email.strip() for email in emails.split(";")]
```

codalab/apps/teams/forms.py
```diff
@@ -89,3 +95,144 @@ def clean(self):
...
+        self.teams_dict = teams_dict
+
+        for team_name in self.teams_dict:
+            if Team.objects.filter(name=team_name):
+                raise ValidationError("A team with that name already exists!")
```
ckcollab: This does a shitload of queries potentially... and it works across all competitions?

Also, we should allow teams that already exist. You should be able to re-run the same CSV over and over without hurting anything?

codalab/apps/teams/models.py
```diff
@@ -283,9 +283,31 @@ def __unicode__(self):
 
 
 class TeamMembership(models.Model):
+    user = models.ForeignKey(settings.AUTH_USER_MODEL)
+    team = models.ForeignKey(Team)
+    is_invitation = models.BooleanField(default=False)
+    is_request = models.BooleanField(default=False)
+    start_date = models.DateTimeField(null=True, blank=True, verbose_name="Start Date (UTC)")
+    end_date = models.DateTimeField(null=True, blank=True, verbose_name="End Date (UTC)")
```
ckcollab: Ah, we made this `end_date` attribute?

codalab/apps/teams/tests/test_organizer_teams.py
```diff
@@ -0,0 +1,177 @@
...
+        # Create a new team with our creator as a member. This should remove us from the first team.
+        resp = self.client.post(reverse('create_org_team', kwargs={'competition_pk': self.competition.pk}),
+                            {"text_members": "test@user.com", "name": "@@@Test@@@", "description": "@@@Test@@@"})
+        assert resp.status_code == 302
+        new_team_to_leave_for = Team.objects.get(name="@@@Test@@@", description="@@@Test@@@")
+        # Check that our creator got removed from the first team, and added to the second.
+        assert self.creator not in new_team.members.all()
+        assert self.creator in new_team_to_leave_for.members.all()
+
+    def test_organizer_teams_csv_file_form(self):
```
ckcollab: is there a test for re-uploading the same CSV?

codalab/apps/teams/forms.py
Tthomas63: We cannot do that unless we change the team name, cancel the operation, or give the user a choice due to names being unique together with competition on teams. And yeah that query should be narrower, on the same competition.

```unique_together = (('name', 'competition'),)```

codalab/apps/teams/models.py
Tthomas63: I don't believe so? I think I moved the fields above the meta declaration be styled like the rest of the classes.

codalab/apps/teams/tests/test_organizer_teams.py
Tthomas63: No there is not. We upload a similar CSV after changing one of the team names from a previous CSV.

codalab/apps/teams/forms.py
Tthomas63: On `self.user_list`? PyCharm was specifying to include it in the constructor since it's something I reference later on. I previously had it in the specific method.

codalab/apps/teams/forms.py
Tthomas63: Ah yes, null and blank are both True on that field.

codalab/apps/teams/forms.py
ckcollab: Yeah, you should probably just do that all in a `clean()` method

codalab/apps/teams/forms.py
ckcollab: Detect that the team exists and augment the members to match what is given?

ckcollab: You can do `update_or_create` ?

codalab/apps/teams/models.py
ckcollab: gotchya

codalab/apps/teams/tests/test_organizer_teams.py
ckcollab: Yeah this is a test case we'll definitely want, re-uploading the same CSV (in case of errors) shouldn't cause problems

codalab/apps/teams/forms.py
Tthomas63: Hmm seems nice having it all in field specific clean methods though.

codalab/apps/teams/forms.py
Tthomas63: Gives too many values to unpack. Ill look at doing this with a list comprehension or something.

codalab/apps/teams/forms.py
Tthomas63: This change caused issues. I don't think it hurts too much to call `save` indiscriminately here

PR #2283, opened 2018-04-29
A teams widgets

ckcollab:


codalab/apps/web/views.py
```diff
@@ -1949,6 +1949,138 @@ def get_success_url(self):
         return reverse('competitions:dumps', kwargs={'competition_pk': dump.competition.pk})
 
 
+class CompetitionWidgetsView(LoginRequiredMixin, DetailView):
+    queryset = models.Competition.objects.all()
+    template_name = 'web/competitions/widgets.html'
+
+    def get_object(self, queryset=None):
```
Tthomas63: Should this possibly be in `dispatch` instead of `get_object`?

codalab/apps/web/views.py
```diff
@@ -1949,6 +1949,138 @@ def get_success_url(self):
...
+        context['phase'] = get_current_phase(self.object)
+        competition = self.object
+
+        if settings.USE_AWS:
+            context['form'] = forms.SubmissionS3UploadForm
+
+        if competition.participants.filter(user__in=[self.request.user]).exists():
+            participant = competition.participants.get(user=self.request.user)
+            if participant.status.codename == models.ParticipantStatus.APPROVED:
+                phase = context['phase']
```
Tthomas63: For readability/style wouldn't it be better to instantiate the phase first, and then assign the context to the phase instead?

```
phase = get_current_phase(self.object)
context['phase'] = phase
```

codalab/apps/web/views.py
```diff
@@ -1949,6 +1949,138 @@ def get_success_url(self):
...
+                    ).select_related('result', 'result__participant')
+                    if leaderboard_entry:
+                        id_of_submission_in_leaderboard = leaderboard_entry[0].result.pk
+                submission_info_list = []
+                for submission in submissions:
+                    try:
+                        default_score = float(submission.get_default_score())
+                    except (TypeError, ValueError):
+                        default_score = '---'
+                    submission_info = {
```
Tthomas63: Should this be surrounded in a Try/Except as well in case one of these fields on submission is None? Shouldn't really matter but what do we have for error checking on the values?

codalab/apps/web/templates/web/widget_iframes/submission.html
```diff
@@ -0,0 +1,185 @@
...
+        }
+        if (auth_window.closed) {
+            window.location.reload()
+        } else {
+            setTimeout(check_auth_loop, 10)
+        }
+    }
+</script>
+</body>
+</html>
```
Tthomas63: No newline at end of file

codalab/apps/web/views.py
Tthomas63: It would make sense to add error handling on grabbing the competition here, but it doesn't look like we require that kind of logic

ckcollab: Yeah Django should take care of object not found problems, but having this check in `dispatch` seems to make more sense than `get_object`, moving it

codalab/apps/web/views.py
ckcollab: Yeah I was lazy about this, nice catch, moving it

codalab/apps/web/views.py
ckcollab: I basically just snatched this code out of the other view, I believe it should handle problems fine. Biggest downside is it's not very DRY but it should be the same as we have currently

codalab/apps/web/templates/web/widget_iframes/submission.html
ckcollab: fixed ty

Issue #2342, opened 2018-07-18
Send Chahub Generic Statistics


Tthomas63:
Things like:
  - Submission Count
  - Participant Count
  - User Count
  - Organizer Count
  - Dataset Count

Needed for: https://github.com/codalab/chahub/pull/47/files

PR #2343, opened 2018-07-18
Send Chahub General Stats

Tthomas63:
Addresses #2342


Tthomas63:
Note: Test failing is from assertion in settings that CHAHUB_BASE_URL ends with a slash.

ckcollab:
tests failing

codalab/apps/chahub/models.py
```diff
@@ -70,7 +70,8 @@ def get_chahub_is_valid(self):
     # -------------------------------------------------------------------------
     def get_chahub_url(self):
         assert settings.CHAHUB_API_URL, "No ChaHub URL given, cannot send details to ChaHub"
-        assert settings.CHAHUB_API_URL.endswith("/"), "ChaHub API url must end with a slash"
+        # Below is handled in settings now
+        # assert settings.CHAHUB_API_URL.endswith("/"), "ChaHub API url must end with a slash"
```
ckcollab: pull?

codalab/apps/chahub/tests/test_chahub_utils.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+        self.phase = CompetitionPhase.objects.create(
+            competition=self.competition,
+            phasenumber=1,
+            start_date=datetime.datetime.now() - datetime.timedelta(days=30),
+        )
+
+
+    @override_settings(CHAHUB_API_URL='https://localhost/')
+    def test_submission_mixin_save_only_retries_once(self):
```
ckcollab: Did you add a new test?

codalab/apps/chahub/utils.py
```diff
@@ -0,0 +1,74 @@
...
+        raise requests.ConnectionError("Chahub is currently offline. Please try again later.")
+
+    url = "{}{}".format(settings.CHAHUB_API_URL, endpoint)
+
+    data = json.dumps(raw_data)
+
+    try:
+        _chahub_send_data(url, data)
+    except requests.ConnectionError:
+        logger.info("Unable to POST data to Chahub. There was an error, please try again.")
```
ckcollab: no need for this, the exception is handled inside `_chahub_send_data`

codalab/apps/chahub/utils.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def _chahub_api_url_check():
```
ckcollab: I think we should just use `assert yadda yadda` instead of this function, now that I see it again?

codalab/apps/chahub/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    try:
+        _chahub_api_url_check()
+    except KeyError:
+        logger.info("CHAHUB_API_URL not defined in settings!")
+        return
+
+    chahub_online = _chahub_online_check()
+
+    if not chahub_online:
```
ckcollab: I don't think we need the temp variable here, can rewrite this function probably like...

```python
def send_to_chahub(endpoint, data):
    assert not settings.CHAHUB_API_URL, "CHAHUB_API_URL not defined in settings!"
    if not _chahub_is_online():
        logger.info("Chahub is currently offline. Cancelling.")
        raise requests.ConnectionError("Chahub is currently offline. Please try again later.")

    url = "{}{}".format(settings.CHAHUB_API_URL, endpoint)
    _chahub_send_data(url, json.dumps(data))
```

`raw_data` should probably just be `data`? We also handle the connection error stuff too many times -- in fact I think that would block the exception from being raised, and no retries would fire

ckcollab: Also, having this function and `_chahub_send_data()` doesn't make a ton of sense, I think? should probably move the contents of that functions into this one?

codalab/codalab/settings/base.py
```diff
@@ -448,6 +448,10 @@ class Base(Settings):
             'task': 'apps.web.tasks.do_chahub_retries',
             'schedule': timedelta(seconds=600),
         },
+        'chahub_general_statistics': {
+            'task': 'apps.web.tasks.send_chahub_general_stats',
+            'schedule': timedelta(seconds=86400),
```
ckcollab: will you put the calculation here so it is more apparent? like, 
```24 * 60 * 60   # 1 day```

File: codalab/apps/web/tasks.py [hunk 7]
```diff
@@ -851,6 +855,32 @@ def do_chahub_retries(limit=None):
...
+def send_chahub_general_stats():
+    if settings.DATABASES.get('default').get('ENGINE') == 'django.db.backends.postgresql_psycopg2':
+        # Only Postgres supports 'distinct', so if we can use the database, if not use some Python Set magic
+        organizer_count = Competition.objects.all().distinct('creator').count()
+    else:
+        users_with_competitions = list(ClUser.objects.filter(competitioninfo_creator__isnull=False))
+        user_set = set(users_with_competitions)
+        # Only unique users that have competitions
+        organizer_count = len(user_set)
+    raw_data = {
```
ckcollab: no need for this temp var

codalab/apps/web/tasks.py
```diff
@@ -851,6 +855,32 @@ def do_chahub_retries(limit=None):
...
+        'competition_count': Competition.objects.count(),
```
ckcollab: count published competitions?

codalab/apps/web/tasks.py
```diff
@@ -851,6 +855,32 @@ def do_chahub_retries(limit=None):
...
+        'dataset_count': OrganizerDataSet.objects.count(),
+        'participant_count': CompetitionParticipant.objects.count(),
```
ckcollab: approved participants?

codalab/apps/chahub/utils.py
```diff
@@ -0,0 +1,59 @@
...
+        raise requests.ConnectionError("Chahub is currently offline. Please try again later.")
+
+    url = "{}{}".format(settings.CHAHUB_API_URL, endpoint)
+
+    data = json.dumps(data)
+
+    _chahub_send_data(url, data)
+
+
+def _chahub_online_check():
```
ckcollab: Huh, is this new? doesn't this already exist somewhere?

codalab/apps/web/tasks.py
Tthomas63: Which variable? Are you saying do all of organizer_count stuff on one line?

Only reason I did not do this is for read-ability and not smashing 2-3 casts in one line

codalab/apps/chahub/utils.py
Tthomas63: Doesn't look like it exists anywhere previously.

ckcollab: Looks like it was plucked from:
```python
@task(queue='site-worker')
def do_chahub_retries(limit=None):
    if not settings.CHAHUB_API_URL:
        return

    logger.info("Checking whether ChaHub is online before sending retries")
    try:
        response = requests.get(settings.CHAHUB_API_URL)
        if response.status_code != 200:
            return
    except requests.exceptions.RequestException:
        # This base exception works for HTTP errors, Connection errors, etc.
        return

    logger.info("ChaHub is online, checking for objects needing to be re-sent to ChaHub")```

Maybe useful to make that function point to this so it's DRY and fix the comment?

File: codalab/apps/chahub/utils.py [hunk 11]
```diff
@@ -0,0 +1,59 @@
...
+    """
+    Does a post request to the specified API endpoint on chahub with the inputted data.
+    :param endpoint: String designating which API endpoint; IE: 'producers/'
+    :param data: Dictionary containing data we are sending away to the endpoint.
+    :return:
+    """
+
+    assert settings.CHAHUB_API_URL
+
+    if not _chahub_online_check():
```
ckcollab: EVERY single send doing online check doesn't make much sense, I don't think

codalab/apps/chahub/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    logger.info("Checking whether ChaHub is online before sending retries")
+    try:
+        response = requests.get(settings.CHAHUB_API_URL)
+        return response.status_code == 200
+    except requests.exceptions.RequestException:
+        # This base exception works for HTTP errors, Connection errors, etc.
+        return False
+
+
+def _chahub_send_data(url, data):
```
ckcollab: Similar problems here, violating DRY i feel like -- if we want these utility functions should pull them out and point to them from both spots

PR #2346, opened 2018-07-26
Remove display: block from .competition-view > code css

### LINKED ISSUE #2345: HTML code block formatting
Currently `<code>` blocks in html are formatted using `display: block` however this is unintuitive as most other platforms use `<pre><code>` to wrap code blocks, and it prevents one from using inline code blocks.

Example (current behaviour):
display-block

Example (removed `display: block` from css):
display-block-disable

I believe the latter should be the default, and the former behaviour can be achieved by wrapping the `<code>` tags in `<pre>` tags.

willprice:
Fixes #2345

### DIFF
--- a/codalab/apps/web/static/less/competitions.less
+++ b/codalab/apps/web/static/less/competitions.less
@@ -8,7 +8,6 @@
     code {
         font-size:@font-size-small;
         margin-bottom:6px;
-        display:block;
         line-height:1.2;
     }
     .organizer {


ckcollab:
Thank you so much for the contribution! This looks like the right spot in LESS for this, were you able to test it locally? Does it maybe break submission description/error output?

willprice:
Hi @ckcollab,
Thanks for the awesome platform, it's great for competition organisers, I'm so glad I only have to worry about writing a scoring program and not a whole platform.

So far in local testing I've not observed any issues with this change. Here's a clip of a stacktrace from a scoring program on my local instance running with the new CSS.

2018-07-26-233009_848x1117_scrot

willprice:
Are there any other places that you can think of that this change might impact? I can check there too.

ckcollab:
I believe that should be it, and if it's not then we can fix it quickly :)

Appreciate the contribution @willprice! If you're interested in V2 development, please watch over here https://github.com/codalab/competitions-v2/ -- can have a lot of say in UX/API and such. By Aug 15th we're going to try and have [competition creation](https://github.com/codalab/competitions-v2/issues/70) flushed out pretty well.

PR #2403, opened 2018-09-13
better retries and tests for sending data to chahub

ckcollab:


codalab/apps/chahub/models.py
```diff
@@ -127,3 +129,8 @@ def save(self, force_to_chahub=False, *args, **kwargs):
 
                     # We save at the beginning, but then again at the end to save our new chahub timestamp and such
                     super(ChaHubSaveMixin, self).save(force_update=True)
+            elif not is_valid and self.chahub_needs_retry:
```
Tthomas63: On what condition are we re-setting chahub_needs_retry?

ckcollab: It should be reset in 2 cases, I believe:

1. We resent the data to Chahub successfully
1. The Chahub data is now invalid, i.e. we tried to send a submission to Codalab but the competition is now unpublished. So, we don't send this invalid submission to Chahub but we unmark `chahub_needs_retry`

ckcollab:
Getting weird test failures here... tests failing on CircleCI pass locally and locally (when running all tests) I get some different fails than CircleCI

ckcollab:
Testing this on `lri-master`

PR #2455, opened 2018-11-10
Chahub send proper updates for removed/deleted items

ckcollab:


ckcollab:
With the "send_participant_counts" changes, this actually should "just work?" Since unpublished competitions aren't included in searches on Chahub and they will be sent now...

PR #2487, opened 2019-01-08
FIX: chart high score doesn't respect sorting

### LINKED ISSUE #2462: Results (low to high)
Hello, 

We're currently running a competition: https://competitions.codalab.org/competitions/20179 

In our competition, a good scoring result has a high positive value (as we are measuring a peak signal to noise ratio, or PSNR).  

On our Results page, the Leaderboard is functioning correctly, with the best scoring submission in position 1 at the top of the Leaderboard.  However, the chart (red plot) is showing the _lowest_ scoring submission for each day.  

How can we reconfigure the red plot to show the _highest_ scoring submission for each day, while keeping the Leaderboard showing the highest score in the first position?

I've tried changing the Leaderboard from Descending to Ascending (Edit->Leaderboard->PSNR->Sorting) but this doesn't work as it puts the highest scoring submission in the last position, and flips the plot upside down.

Many thanks for any help!

MarcCote:
At the moment, high scores displayed in the chart are wrong, i.e. they don't respect the sorting specified for the metric.

This PR fixes #2462.

As an example, see https://competitions.codalab.org/competitions/20865#results where all the high in the chart are 0 while in the leaderboard table we clearly see submissions with high scores in the hundreds.

image

### DIFF
--- a/codalab/apps/web/views.py
+++ b/codalab/apps/web/views.py
@@ -550,9 +550,9 @@ class CompetitionDetailView(DetailView):
                     qs = SubmissionScore.objects.filter(result__phase__competition=competition, scoredef=score_def)
                     qs = qs.extra({'day': truncate_date}).values('day')
                     if score_def.sorting == 'asc':
-                        best_value = Max('value')
-                    else:
                         best_value = Min('value')
+                    else:
+                        best_value = Max('value')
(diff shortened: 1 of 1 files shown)


ckcollab:
Thank you very much!

MarcCote:
When will this fix be deployed?

PR #2515, opened 2019-02-18
Force best to leaderboard another test for smaller differences

ckcollab:


ckcollab:
Hope this doesn't break force best submissions..! :)

PR #2544, opened 2019-03-22
2541 congested mode

gibsonbailey:


codalab/apps/web/tasks.py
```diff
@@ -816,9 +816,19 @@ def send_mass_email(competition_pk, body=None, subject=None, from_email=None, to
     text = render_to_string("emails/notifications/participation_organizer_direct_email.txt", context)
     html = render_to_string("emails/notifications/participation_organizer_direct_email.html", context)
 
-    mail_tuples = ((subject, text, html, from_email, [e]) for e in to_emails)
+    message_info = []
+    site = Site.objects.get_current()
 
-    _send_mass_html_mail(mail_tuples)
```
ckcollab: ah, does this also fix a user context problem?

By the way, we recently had a problem where emailing participants count > 100 seems to not be sending for some reason.. if you setup a free SendGrid account you can send a couple hundred emails to see if you can replicate this recent problem?

codalab/apps/web/templates/web/competitions/_submit_results_page.html
```diff
@@ -25,6 +25,12 @@ <h4>Phase description</h4>
         <p><b>Max submissions total: </b> {{ phase.max_submissions }}</p>
     </div>
 
+    {% if submission_threshold_met %}
+        <div class="alert alert-danger text-center">
+            <h2>We already have too many submissions running. Please come back in a few minutes.</h2>
```
ckcollab: Could you maybe say:

> _Warning: The compute queues are congested, please wait for existing jobs to process before submitting more. Thanks for your patience!_

codalab/apps/web/views.py
```diff
@@ -734,6 +735,17 @@ def get_context_data(self, **kwargs):
             context['last_submission_organization_or_affiliation'] = last_submission.organization_or_affiliation
         except ObjectDoesNotExist:
             pass
+
+        jobs_today = Job.objects.filter(created__year=datetime.today().year,
```
ckcollab: nitpick, should be formatted like this to follow pep8:

```python
        jobs_today = Job.objects.filter(
            created__year=datetime.today().year,
            created__day=datetime.today().day,
            created__month=datetime.today().month
        )
```

If the existing code is formatted wrong it's ok to fix that, all "new code" we should try to style nicely, or if we see something messed up fix it as we go.

codalab/apps/web/tasks.py
ckcollab: Or: maybe something was branched wrong?

ckcollab:
Also, tests failing atm

codalab/apps/web/views.py
gibsonbailey: Sounds great to me.

codalab/apps/web/templates/web/competitions/_submit_results_page.html
gibsonbailey: Done! Looks much more professional.

codalab/apps/web/tasks.py
gibsonbailey: I believe this was a branching problem. I purged my local workspace of old commits, so this problem does not keep appearing.

PR #2548, opened 2019-03-29
2541 congested mode

gibsonbailey:


gibsonbailey:
Screen Shot 2019-03-29 at 11 43 01 AM
Preview of congestion message

codalab/apps/health/models.py
```diff
@@ -5,3 +5,4 @@ class HealthSettings(models.Model):
     """Base Health Settings Model. Counts the amount of jobs ready to process."""
     emails = models.TextField(null=True, blank=True)
     threshold = models.PositiveIntegerField(default=25, null=True, blank=True)
+    congestion_threshold = models.PositiveIntegerField(default=100, null=True, blank=True)
```
ckcollab: This shouldn't be allowed to be null/blank, I think? The default should protect that case though.

PR #2624, opened 2019-06-11
Worker monitor

ckcollab:


ckcollab:
I pushed the new worker I made as well (code for ref: https://github.com/codalab/competitions-v1-compute-worker/pull/5/) -- it's already referenced in `docker-compose.yaml`

codalab/apps/health/templates/health/worker_list.html
```diff
@@ -0,0 +1,235 @@
...
+        var ctx = document.getElementById('myChart');
+
+        var the_data = {{ json_data|safe }}
+
+        the_data.forEach(function(task_list){
+            task_list.forEach(function(task){
+                if(task.data === undefined) {
+                    return
+                }
+                for(var i = 0; i < task.data.length; i++) {
```
jimmykodes: 90% sure this should work the same as this for loop, but cleaner
```javascript
task.data = task.data.map(function(time) {
    new moment(time)
})
```

docker-compose.yml
```diff
@@ -198,12 +198,13 @@ services:
     container_name: worker_site
       
   worker_compute:
-    image: codalab/competitions-v1-compute-worker:testing-realtime2
+    image: codalab/competitions-v1-nvidia-worker:reporter
```
Tthomas63: Are we pushing this to latest? May want to have this point there

ckcollab: Good call

codalab/apps/health/models.py
```diff
@@ -1,8 +1,63 @@
...
+            self.worker.unique_id,
+            "up" if self.up else "down",
+        )
+
+
+class TaskMetadata(models.Model):
+    worker = models.ForeignKey(Worker, null=True, blank=True, related_name="tasks")
+    submission = models.ForeignKey('web.CompetitionSubmission', related_name="tasks")
+    is_predicting = models.BooleanField(default=True)
+    is_scoring = models.BooleanField(default=False)
```
jimmykodes: do we need _both_ `is_scoring` and `is_predicting` if one is inherently the opposite of the other?

codalab/apps/web/tasks.py
```diff
@@ -605,7 +608,99 @@ def __init__(self, submission, inner_exception):
...
+        gpus,
+        vhost,
+    ))
+
+    if vhost != '/':
+        queue = Queue.objects.get(vhost=vhost)
+    else:
+        queue = None
+
+    worker = Worker.objects.update_or_create(
```
jimmykodes: Do we need to set this to a variable?

codalab/apps/web/tasks.py
```diff
@@ -605,7 +608,99 @@ def __init__(self, submission, inner_exception):
...
+            "harddrive_gb": harddrive_gb,
+            "gpus": gpus,
+            "queue": queue
+        },
+    )
+
+
+@app.task(queue='submission-updates')
+def worker_job_started(worker_id, submission_secret, is_scoring):
+    assert worker_id, "WORKER_ID cannot be null, received: {}".format(worker_id)
```
jimmykodes: General stylistic question: Does it make sense to have asserts in business logic? do we want it to actually throw an error? or should this be something like:
```python
if not worker_id:
    logger.warning('WORKER_ID cannot be null...')
    return
```

codalab/apps/web/tasks.py
```diff
@@ -605,7 +608,99 @@ def __init__(self, submission, inner_exception):
...
+            "end": timezone.now(),
+        }
+    )
+
+
+@app.task(queue='submission-updates')
+def worker_keep_alive(worker_id):
+    assert worker_id, "WORKER_ID cannot be null, received: {}".format(worker_id)
+    logger.info("Keep alive received for worker id = {}".format(worker_id))
+    worker = Worker.objects.get_or_create(unique_id=worker_id)[0]
```
jimmykodes: To keep w/ general django convention, shouldn't we use tuple unpacking here?
Or what about update_or_create instead?

PR #2633, opened 2019-07-03
2199 newsletter

BrendanMyers88:
Overview
- [x] Setup mailing list for Codalab

Backend rough specs
- [x] MAILING_LIST = tbd (mailchimp, sendgrid, etc.)

- [x] We need to add new attribute to user models like newsletter_opt_in
- [x] Then we'll need to on sign up of new people have a "newsletter opt in" checkbox
- [x] We can send each new sign up to MAILING_LIST
- [ ] Current users do not have an attribute related to "opting in" to a newsletter… seems like a dick move to put people on the list automatically? where did we land with isabelle on that detail? cool to auto-add people if they have like organizer_status_updates set to true?

- [x] If we have existing users, send them to MAILING_LIST
- [ ] Manually send out June newsletter on MAILING_LIST


codalab/apps/authenz/models.py
```diff
@@ -32,4 +41,57 @@ class ClUser(auth_models.AbstractUser):
...
     rabbitmq_password = models.CharField(max_length=36, null=True, blank=True)
 
+    def save(self, *args, **kwargs):
+        if self.newsletter_opt_in and self.email and self.is_active:
+            data = {
+                "email_address": self.email,
+                "status": "subscribed",
+            }
+
+            user_hash = hashlib.md5(str.lower(self.email.encode()))
```
jimmykodes: Will this work? or do you have to explicitly call it with `str`?
```python
user_hash = hashlib.md5(self.email.encode().lower())
```

codalab/apps/authenz/models.py
```diff
@@ -32,4 +41,57 @@ class ClUser(auth_models.AbstractUser):
     rabbitmq_username = models.CharField(max_length=36, null=True, blank=True)
     rabbitmq_password = models.CharField(max_length=36, null=True, blank=True)
 
+    def save(self, *args, **kwargs):
```
jimmykodes: We should check to make sure we have all the mailchimp settings set before we try and make mailchimp requests.
We also need to catch exceptions from `requests`.  I believe `requests.exceptions.RequestException` is the catch-all exception. 
We should probably do something like the `chahub_needs_retry` logic so we make sure updates actually get through to mailchimp, _especially_ in the case of an unsubscribe.

codalab/apps/authenz/models.py
```diff
@@ -32,4 +41,57 @@ class ClUser(auth_models.AbstractUser):
...
+                    settings.MAILCHIMP_MEMBERS_ENDPOINT_NEWSLETTER,
+                    auth=("", settings.MAILCHIMP_API_KEY),
+                    data=json.dumps(data)
+                )
+
+            # If there is no NewsletterUser for this user, create one
+            if not NewsletterUser.objects.filter(email=self.email).exists():
+                NewsletterUser.objects.create(email=self.email)
+
+        elif not self.email and self.newsletter_opt_in:
```
jimmykodes: Can check this first, as a standalone if statement. then, after this `if`, you can calculate your `user_hash` outside of the next `if...else` and then you only have to calc it once, so you can make a small DRY reduction there.

codalab/apps/authenz/models.py
```diff
@@ -32,4 +41,57 @@ class ClUser(auth_models.AbstractUser):
...
+
+            # Update user in Mailchimp to Unsubscribed
+            requests.patch(
+                settings.MAILCHIMP_MEMBERS_ENDPOINT_NEWSLETTER + '/' + user_hash.hexdigest(),
+                auth=("", settings.MAILCHIMP_API_KEY),
+                data=json.dumps(data)
+            )
+
+            # Remove the user from the NewsletterUser model
+            if NewsletterUser.objects.filter(email=self.email).exists():
```
jimmykodes: Should be able to do
```python
NewsletterUser.objects.filter(email=self.email).delete()
```
which will delete the objects the queryset returns, if any exist, and do nothing if the qs is empty.

codalab/apps/authenz/models.py
```diff
@@ -32,4 +41,57 @@ class ClUser(auth_models.AbstractUser):
...
+            # If no user is found in mailchimp, create a new mailing list user
+            if not update_user.ok:
+                requests.post(
+                    settings.MAILCHIMP_MEMBERS_ENDPOINT_NEWSLETTER,
+                    auth=("", settings.MAILCHIMP_API_KEY),
+                    data=json.dumps(data)
+                )
+
+            # If there is no NewsletterUser for this user, create one
+            if not NewsletterUser.objects.filter(email=self.email).exists():
```
jimmykodes: Does this version of django have `update_or_create()` or `get_or_create()`? this may be a good time to use one of them, if they are available.

codalab/apps/newsletter/scripts/send_emails_to_mailchimp.py
```diff
@@ -0,0 +1,27 @@
...
+            "status": "subscribed",
+        }
+
+        requests.patch(
+            settings.MAILCHIMP_MEMBERS_ENDPOINT_ALL,
+            auth=("", settings.MAILCHIMP_API_KEY),
+            data=json.dumps(data)
+        )
+
+        requests.post(
```
jimmykodes: why do we patch and then post the same data?

codalab/apps/newsletter/scripts/send_emails_to_mailchimp.py
```diff
@@ -0,0 +1,27 @@
+import json
+import requests
+
+from django.contrib.auth import get_user_model
+from codalab import settings
+
+User = get_user_model()
+
+
+def run():
```
jimmykodes: need to make sure mailchimp settings exist. Also, catch request exceptions here too

codalab/apps/newsletter/templates/newsletter/unsubscribe_email.txt
```diff
@@ -0,0 +1,5 @@
+{% extends 'newsletter/base_email.txt' %}
+
+{% block content %}
+We're sorry to see you go! You have unsubscribed from the Codalab newsletter. If there was an issue on our end, please let us know and we'll do our best to get it resolved! If you would like to sign up again, please visit http://localhost:8000/newsletter/signup
```
jimmykodes: need to replace this hardcoded URL

codalab/apps/newsletter/tests/test_newsletter_opt_in.py
```diff
@@ -0,0 +1,133 @@
...
+from django.test import TestCase
+from django.contrib.auth import get_user_model
+
+from apps.newsletter.models import NewsletterUser
+from codalab import settings
+
+User = get_user_model()
+
+
+class NewsletterOptIn(TestCase):
```
jimmykodes: We should definitely be `mock.patch`ing mailchimp here.

codalab/apps/authenz/models.py
```diff
@@ -32,4 +41,57 @@ class ClUser(auth_models.AbstractUser):
...
+
+        elif not self.newsletter_opt_in and self.email:
+            data = {
+                "status": "unsubscribed",
+            }
+
+            user_hash = hashlib.md5(str.lower(self.email.encode()))
+
+            # Update user in Mailchimp to Unsubscribed
+            requests.patch(
```
jimmykodes: Do we need to do this here if the pre_delete signal is also kicking off the unsubscribe process?

codalab/apps/newsletter/models.py
```diff
@@ -0,0 +1,33 @@
...
+class NewsletterUser(models.Model):
+    email = models.EmailField(null=True, blank=True, default=None)
+    date_added = models.DateTimeField(auto_now_add=True)
+
+    def __str__(self):
+        return self.email
+
+
+@receiver(pre_delete, sender=NewsletterUser)
+def newsletteruser_unsubscribe(instance, **kwargs):
```
jimmykodes: also check mailchimp settings and catch request errors.

codalab/apps/authenz/models.py
jimmykodes: So thinking about this a little more completely, I think we should move this business logic to the `NewsletterUser` model. That model should have like a `subscribe` and `unsubscribe` method, which will calculate the hash itself (maybe even store the hash as a field on that model? or maybe a `@property`?) as well as handle request exceptions and all that. This will clean up the `User` model `save` logic quite a lot. Would probably reduce a lot of DRY violation throughout the code as well.

codalab/apps/newsletter/templates/newsletter/signup_email.txt
```diff
@@ -0,0 +1,5 @@
+{% extends 'newsletter/base_email.txt' %}
+
+{% block content %}
+Welcome to the Codalab newsletter. If you would like to unsubscribe, please visit http://localhost:8000/newsletter/unsubscribe
```
jimmykodes: Need to fix hardcoded URL

codalab/apps/newsletter/tests/test_newsletter_opt_in.py
```diff
@@ -0,0 +1,133 @@
...
+            )
+
+            if delete_status.ok:
+                print('Test user has been deleted from the Mailchimp mailing list')
+            else:
+                print('Could not delete this user from the Mailchimp mailing list')
+        else:
+            print('User was not found on the Mailchimp mailing list')
+
+    def test_create_user_newsletter_opt_in_true(self):
```
jimmykodes: All these tests need better names so we know what each one is supposed to be testing. Like
```python
def test_user_creation_makes_newsletter_user_on_opt_in(self):
```
or something like that

codalab/apps/newsletter/models.py
```diff
@@ -0,0 +1,33 @@
...
+import requests
+from django.db import models
+from django.db.models.signals import pre_delete
+from django.dispatch import receiver
+
+from codalab import settings
+
+
+class NewsletterUser(models.Model):
+    email = models.EmailField(null=True, blank=True, default=None)
```
jimmykodes: We probably don't want this to be nullable? We can't have a newsletter user without an email. 
I think we also want this to be `unique=True` so that we only create one object per email address.

codalab/apps/newsletter/models.py
```diff
@@ -0,0 +1,33 @@
...
+
+import requests
+from django.db import models
+from django.db.models.signals import pre_delete
+from django.dispatch import receiver
+
+from codalab import settings
+
+
+class NewsletterUser(models.Model):
```
jimmykodes: Maybe a better name would be `NewsletterSubscription` since this isn't really a user object?

codalab/apps/authenz/models.py
BrendanMyers88: User hash has been added to its own self-contained function in the model

codalab/apps/authenz/models.py
BrendanMyers88: Fixed in latest push

codalab/apps/authenz/models.py
BrendanMyers88: Resolved in latest push

codalab/apps/authenz/models.py
BrendanMyers88: Changed the way the logic works to unsubscribe but not delete NewsletterSubscriptions

codalab/apps/newsletter/scripts/send_emails_to_mailchimp.py
BrendanMyers88: Mailchimp suggests this in their documentation. First it checks if the user is in their API and if not, we'll add them. Not ideal but that's how they've got it set up

codalab/apps/newsletter/templates/newsletter/unsubscribe_email.txt
BrendanMyers88: Fixed in latest push

codalab/apps/newsletter/tests/test_newsletter_opt_in.py
BrendanMyers88: Mock added to the tests

codalab/apps/authenz/models.py
BrendanMyers88: Changed logic in latest push

codalab/apps/newsletter/templates/newsletter/signup_email.txt
BrendanMyers88: Fixed in latest push

codalab/apps/newsletter/tests/test_newsletter_opt_in.py
BrendanMyers88: renamed tests to be more clear

codalab/apps/newsletter/models.py
BrendanMyers88: Updated in latest push. I had to make add some logic to the is_valid() function for the Unsubscribe form because the `unique=True` didn't like that we were passing non-unique emails to the unsub form.

codalab/apps/newsletter/models.py
BrendanMyers88: Updated in latest push

codalab/apps/newsletter/scripts/send_emails_to_mailchimp.py
jimmykodes: So, should we only `post` if `patch` returns a non-ok response?

BrendanMyers88: Fixed in latest push

codalab/apps/newsletter/forms.py
```diff
@@ -0,0 +1,29 @@
...
+        model = NewsletterSubscription
+        fields = ['email']
+
+        def clean_email(self):
+            email = self.cleaned_data.get()
+
+            return email
+
+
+class NewsletterSubscriptionUnsubscribeForm(forms.ModelForm):
```
jimmykodes: We should probably use `forms.Form` instead of a `ModelForm`, then we don't need to override anything about unique values.

codalab/apps/newsletter/views.py
```diff
@@ -0,0 +1,110 @@
...
+            requests.post(
+                settings.MAILCHIMP_MEMBERS_ENDPOINT_NEWSLETTER,
+                auth=("", settings.MAILCHIMP_API_KEY),
+                data=json.dumps(data)
+            )
+
+            instance.save()
+            messages.success(request, 'You have been added to the Codalab newsletter',
+                             'alert alert-success alert-dismissible')
+            subject = "Thank you for joining the Codalab newsletter"
```
jimmykodes: Should put this logic into a small helper function to reduce DRY, similar to:
https://github.com/codalab/codalab-competitions/blob/49ce6a472b6ab04511c04636b59db4df8fda9ef0/codalab/apps/api/views/competition_views.py#L203

codalab/apps/newsletter/forms.py
BrendanMyers88: Updated in latest push

codalab/apps/newsletter/views.py
BrendanMyers88: Added a helper function in latest push for `_send_mail()`

codalab/apps/newsletter/models.py
```diff
@@ -0,0 +1,103 @@
...
+                response = requests.post(
+                    settings.MAILCHIMP_MEMBERS_ENDPOINT_NEWSLETTER,
+                    auth=("", settings.MAILCHIMP_API_KEY),
+                    data=json.dumps(data)
+                )
+
+                self.needs_retry = not response.ok
+                self.save()
+            else:
+                print('POST FAILED, NEEDS RETRY')
```
jimmykodes: A bunch of print statements in here we should probably pull

codalab/apps/newsletter/tasks.py
```diff
@@ -0,0 +1,11 @@
+from celery import task
+
+from apps.newsletter.models import NewsletterSubscription
+
+
+@task(queue='site-worker', soft_time_limit=60 * 5)
+def retry_mailing_list():
```
jimmykodes: Just thought of things we need to do for this. 

We need to check and make sure we have mailchimp settings. Even though this is done in the subscribe/unsubscribe methods, we should do it here before we start hitting the database.

Also, if we do have mailchimp settings, we should fire off a `get` request to mailchimp and make sure we get an `ok` status back so we don't make a bunch of `patch`/`post` requests that are just going to error out.

codalab/apps/newsletter/tests/test_newsletter_opt_in.py
jimmykodes: should also add a few more tests:
```python 
def test_needs_retry_on_patch_fail(self):
    self.user.newsletter_opt_in = True
    # clear the detail so we don't send a post request
    mock_post, mock_patch = self.mock_save_user(patch_status_code=400, post_status_code=400, detail='')
    assert mock_patch.called
    assert not mock_post.called
    assert NewsletterSubscription.objects.filter(email=self.user.email).exists()
    newsletter_user = NewsletterSubscription.objects.get(email=self.user.email)
    assert newsletter_user.needs_retry
    assert newsletter_user.subscription_active

def test_needs_retry_on_patch_fail(self):
    self.user.newsletter_opt_in = True
    mock_post, mock_patch = self.mock_save_user(patch_status_code=400, post_status_code=400)
    assert mock_patch.called
    assert mock_post.called
    assert NewsletterSubscription.objects.filter(email=self.user.email).exists()
    newsletter_user = NewsletterSubscription.objects.get(email=self.user.email)
    assert newsletter_user.needs_retry
    assert newsletter_user.subscription_active

def test_needs_retry_on_unsub_fail(self):
    self.user.newsletter_opt_in = False
    mock_post, mock_patch = self.mock_save_user(patch_status_code=400, post_status_code=400)
    assert mock_patch.called
    assert mock_post.called
    assert NewsletterSubscription.objects.filter(email=self.user.email).exists()
    newsletter_user = NewsletterSubscription.objects.get(email=self.user.email)
    assert newsletter_user.needs_retry
    assert not newsletter_user.subscription_active
```

I think these all should work. but I didn't wanna take the time to set shit up on my end and actually run the tests, but I think the logic is solid.

jimmykodes: Also, looks like there are a couple tests failing on circleCI

jimmykodes:
Looks good to me. Once Tyler gives the green light, can merge

PR #2659, opened 2019-07-24
Removes Newsletter functionality if no api key assigned

BrendanMyers88:
Removes the ability for users to access the newsletter signup/unsubscribe if there is no API key given. This is designed to prevent confusion in the case where someone hosting a version of CodaLab doesn't want to send newsletters.


codalab/apps/authenz/forms.py
```diff
@@ -7,7 +8,8 @@ class CodalabSignupForm(forms.Form):
     """
     team_name = forms.CharField(max_length=64, required=False)
     contact_email = forms.EmailField(required=False)
-    newsletter_opt_in = forms.BooleanField(initial=False, required=False)
+    if settings.MAILCHIMP_API_KEY:
+        newsletter_opt_in = forms.BooleanField(initial=False, required=False)
```
ckcollab: Should probably not do it like this, instead make it default to False and not required, hide the field on the frontend?

codalab/apps/newsletter/views.py
```diff
@@ -25,74 +25,80 @@ def _send_mail(context_data, from_email=None, html_file=None, text_file=None, su
 
 
 def newsletter_signup(request):
-    form = NewsletterSubscriptionSignUpForm(request.POST or None)
+    if settings.MAILCHIMP_API_KEY:
```
ckcollab: I don't think we need to do this check on these views, can just hide the ability to see/hit newsletter related stuff and have nice default behavior if no api key

codalab/apps/web/templates/account/signup.html
```diff
@@ -38,6 +38,13 @@
                 <input id="id_confirmation_key" name="confirmation_key" type="hidden">
             </div>
 
+            {% if settings.MAILCHIMP_API_URL %}
```
ckcollab: purrrfect, can have an `{% else %}` with...

```html
<input type='hidden' id="id_newsletter_opt_in" value="">
```

Or what-have-you, above may not work exactly correctly but I believe empty value should evaluate to false

codalab/apps/authenz/forms.py
BrendanMyers88: All fields are now hidden on the front end via context processor.

codalab/apps/newsletter/views.py
BrendanMyers88: Added a default to 404 if no api key is set, otherwise we'll go to the correct form view.

codalab/apps/web/templates/account/signup.html
BrendanMyers88: Worked great! Thanks!

codalab/apps/newsletter/views.py
```diff
@@ -25,6 +25,10 @@ def _send_mail(context_data, from_email=None, html_file=None, text_file=None, su
 
 
 def newsletter_signup(request):
+    if not settings.MAILCHIMP_API_KEY:
+        template = "newsletter/404.html"
```
ckcollab: can do `raise Http404()` here and it'll automatically go to the existing 404 template -- did we need a new one?

codalab/apps/newsletter/views.py
```diff
@@ -50,19 +54,25 @@ def newsletter_signup(request):
...
-        messages.warning(request, 'Please use a valid email address to subscribe to the newsletter',
-                         'alert alert-warning alert-dismissible')
-
     context = {
         'form': form,
     }
 
     template = "newsletter/signup.html"
+
+    storage = messages.get_messages(request)
```
ckcollab: whoa what is this doing?!

codalab/codalab/context_processors.py
```diff
@@ -22,4 +27,5 @@ def common_settings(request):
         'is_dev': codalab_settings.IS_DEV,
         'USE_AWS': codalab_settings.USE_AWS,
         'CODALAB_SITE_DOMAIN': codalab_settings.CODALAB_SITE_DOMAIN,
+        'USE_MAILCHIMP': use_mailchimp,
```
ckcollab: instead of the temp var `use_mailchimp` can do..

```python
'USE_MAILCHIMP': bool(settings.MAILCHIMP_API_KEY)
```

(ha! makes me remember 'vool' from like 11 years ago :D)

codalab/apps/newsletter/views.py
BrendanMyers88: Raises Http404() in latest push

codalab/apps/newsletter/views.py
BrendanMyers88: Removes stale message from the cache. I removed the unnecessary loop in the latest push.

codalab/codalab/context_processors.py
BrendanMyers88: Added vool() to latest push, Such a clean way to do this!

PR #2671, opened 2019-08-07
is_active field added to phase serializer

gibsonbailey:


codalab/apps/api/serializers.py
```diff
@@ -54,10 +54,51 @@ class Meta:
 
 class PhaseSerial(serializers.ModelSerializer):
     start_date = serializers.DateField(format='%Y-%m-%d')
+    is_active = serializers.Field()
 
     class Meta:
         model = webmodels.CompetitionPhase
-        read_only_fields = ['datasets']
+        fields = (
```
Tthomas63: I'm not sure we want to expose all of these fields on the API. I might've misspoke yesterday and said to do this, but we probably only want metadata/descriptor fields about the phase, not files, etc.

gibsonbailey: Only useful fields exposed now.

Issue #2606, opened 2019-05-09
After I delete a competition in "Competitions I'm Running", I can still see it in "Competitions I'm In".


hqqasw:
*** After I delete a competition in "Competitions I'm Running", I can still see it in "Competitions I'm In". ***

I am an organizer and I found that 
after I delete a competition in "Competitions I'm Running", I can still see it in "Competitions I'm In".
And when I click that competition in "Competitions I'm In", it would show a bug-reporte page.
------------------------------------------------------------------

Didayolo:
Solved in develop branch.

PR #2693, opened 2019-09-13
Bug fix: add related manager to CompetitionParticipant for soft delet…

Didayolo:
Add related manager to CompetitionParticipant for soft delete. 

When deleting a competition it is now deleted from "My competitions".

Related issue: After I delete a competition in "Competitions I'm Running", I can still see it in "Competitions I'm In". #2606


Didayolo:
I just had this issue. We should merge this.

codalab/apps/web/models.py
```diff
@@ -1244,6 +1249,13 @@ def sortkey(x):
         return results
 
 
+class CompetitionParticipantManager(models.Manager):
```
ckcollab: this appears unused? should be deleted?

Didayolo: I think so, can't find this class anywhere else in the code.

PR #2700, opened 2019-10-03
Chahub backend for AutoDL

jimmykodes:


codalab/apps/chahub/context_processors.py
```diff
@@ -0,0 +1,6 @@
+from allauth.socialaccount import providers
+
+
+def socialaccount(request):
+    ctx = {'providers': providers.registry.get_list()}
+    return dict(socialaccount=ctx)
```
ckcollab: This looks a little confusing, could return a dictionary with social account key, with provider key inside it?

codalab/apps/chahub/models.py
```diff
@@ -56,15 +56,14 @@ def get_chahub_data(self):
         raise NotImplementedError()
 
     def get_chahub_is_valid(self):
-        """Override this to validate the specifc model before it's sent
+        """Override this to validate the specific model before it's sent
```
ckcollab: Ty :)

codalab/apps/chahub/context_processors.py
jimmykodes: I can change the way this function looks, but I returned the dictionary of this specific shape since frontend templates were already looking for this context processor (copied from v1.5) but the new version of django allauth removed the context processors natively, so I reconstructed it. Pretty sure that was a confusing sentence... anyway, this maintains the greatest similarity between 1.5 and 1.8. I can definitely change it, it just involves changes to template files, too.

jimmykodes: My first comment didn't make any sense... anyway, yes, I can restructure this to look less... weird.. I just grabbed the context processor code straight from django allauth's old code (they deprecated it between the version on 1.5 and the version on 1.8)

PR #2717, opened 2019-11-16
Add is active to phase serializer master

gibsonbailey:
The purpose of this PR is to expedite the exposure of the `is_active` field on the phase serializer for Chagrade's use.


codalab/apps/api/serializers.py
```diff
@@ -54,10 +54,29 @@ class Meta:
 
 class PhaseSerial(serializers.ModelSerializer):
     start_date = serializers.DateField(format='%Y-%m-%d')
+    is_active = serializers.Field()
```
ckcollab: is this part necessary? I don't believe you need this here?

gibsonbailey: The reason I do this is because `is_active` is a property on the `CompetitionPhase` model. In the DRF docs [here](https://www.django-rest-framework.org/api-guide/serializers/#specifying-fields-explicitly).

Screen Shot 2019-11-18 at 9 44 38 AM

ckcollab: for future reference: this broke without `is_active = serializers.Field()`

PR #2725, opened 2019-12-02
Charepo data sending

ckcollab:


codalab/apps/chahub/utils.py
```diff
@@ -29,6 +29,9 @@ def send_to_chahub(endpoint, data, update=False):
                 'X-CHAHUB-API-KEY': settings.CHAHUB_API_KEY,
             }
         }
-        return requests.post(**kwargs)
+        if update:
+            return requests.patch(**kwargs)
```
jimmykodes: chahub doesn't take patch requests, everything is done through post.

codalab/apps/chahub/utils.py
```diff
@@ -29,6 +29,9 @@ def send_to_chahub(endpoint, data, update=False):
...
+        else:
+            return requests.post(**kwargs)
```
jimmykodes: doesn't look like your data is a list, so I think that's gotta be the issue.

ckcollab: good catch

codalab/apps/chahub/utils.py
ckcollab: cool will pull this

codalab/apps/web/templates/web/my/datasets.html
```diff
@@ -50,15 +51,21 @@
                                             <td>{{ dataset.key }}</td>
                                             <td>{{ dataset.description }}</td>
                                             <td>{{ dataset.type }}</td>
+                                            <td>{{ dataset.is_public|default:"" }}</td>
                                             <td>
-                                                <a href="{% url 'datasets_download' dataset_key=dataset.key %}" class="button btn btn-default btn-sm">Download</a>
+                                                <a href="{% url 'datasets_download' dataset_key=dataset.key %}" class="button btn btn-primary btn-sm">Download</a>
+                                                {% if dataset.is_public %}
+                                                    <a href="{% url 'my_datasets_publish' dataset_key=dataset.key %}" class="button btn btn-default btn-sm publish">Unpublish</a>
```
jimmykodes: Should probably add some kind of `if(confirm('You sure?'))` to this. So at least we can say "we warned you" if people publish a dataset and shoot their competition in the foot.

codalab/apps/web/templates/web/my/datasets.html
```diff
@@ -50,15 +51,21 @@
                                             <td>{{ dataset.key }}</td>
                                             <td>{{ dataset.description }}</td>
                                             <td>{{ dataset.type }}</td>
+                                            <td>{{ dataset.is_public|default:"" }}</td>
```
jimmykodes: Does this correctly turn a bool into a string? I think we had to do something with `yesno` to make this work right on v2. Something like:
```
{{dataset.is_public|yesno:"True,False"}}
```

ckcollab: the existing code prints `"True"` or nothing if false

codalab/apps/web/templates/web/competitions/edit.html
```diff
@@ -246,7 +246,7 @@ <h3 class="panel-title">
                 </li>
             {% endfor %}
         </ul>
-        <button type="button" class="medium blue button add-phase btn btn-secondary pull-right">Add Phase</button>
+{#        <button type="button" class="medium blue button add-phase btn btn-secondary pull-right">Add Phase</button>#}
```
jimmykodes: What are we doing with this?

codalab/apps/web/templates/web/my/datasets.html
jimmykodes: Cool. Maybe I'm misremembering the issue then. 👍

codalab/apps/web/templates/web/my/datasets.html
ckcollab: added

codalab/apps/web/templates/web/competitions/edit.html
ckcollab: I think this is a remnant from something else, annotate is saying this was edited in 2018 ... ?

codalab/apps/web/templates/web/my/submissions.html
```diff
@@ -174,15 +174,15 @@ <h4>Submissions</h4>
 
         <nav aria-label="Page navigation example">
             <ul class="pagination">
-                    {% if submission_info_list.has_previous %}
-                        <li><a href="?page=1{% if selected_phase %}&phase={{ selected_phase.pk }}{% endif %}">&laquo; first</a></li>
-                        <li><a href="?page={{ submission_info_list.previous_page_number }}{% if selected_phase %}&phase={{ selected_phase.pk }}{% endif %}">previous</a></li>
-                    {% endif %}
+                {% if submission_info_list.has_previous %}
```
ckcollab: looks like just indentation changed, so this should be fine

codalab/apps/web/templates/web/competitions/edit.html
ckcollab: this shouldn't be showing up any more?

codalab/apps/web/models.py
```diff
@@ -1439,6 +1446,8 @@ def save(self, ignore_submission_limits=False, *args, **kwargs):
 
         # only at save on object creation should it be submitted
         if not self.pk:
+            if self.participant.user.username == 'chagrade_bot':
+                ignore_submission_limits = True
```
jimmykodes: wonder why this is showing up in here...

codalab/apps/web/models.py
```diff
@@ -2315,11 +2324,17 @@ class OrganizerDataSet(models.Model):
     sub_data_files = models.ManyToManyField('OrganizerDataSet', null=True, blank=True, verbose_name="Bundle of data files")
     uploaded_by = models.ForeignKey(settings.AUTH_USER_MODEL)
     key = UUIDField(version=4)
+    is_public = models.BooleanField(default=False)
+    created_when = models.DateTimeField(null=True, blank=True)
```
jimmykodes: should we do `default=now` instead of checking in the save?

codalab/apps/web/models.py
ckcollab: :( not sure, but we certainly want that...

codalab/apps/web/models.py
ckcollab: I do not want to set default on old models so I'm a bit hesitant to do that. When we do this migration, I think it will set ALL old datasets `created_when` to `now()`

codalab/apps/web/views.py
```diff
@@ -1580,6 +1583,19 @@ def download_dataset(request, dataset_key):
         return HttpResponse(msg, status=400, content_type='text/plain')
 
 
+def publish_dataset(request, dataset_key):
```
jimmykodes: maybe "toggle_dataset_publish" or something like that? since it will also _un_ publish?

ckcollab: ✅ yeah i was thinking the same

codalab/apps/web/models.py
jimmykodes: if those old instances ever get updated, their `created_when` will get set to `now` also, but yea, probably fewer will be affected that way.

ckcollab: yeah that is true... maybe I could do some clever stuff and get "last modified" from the file or something? =\ probably not worth it

jimmykodes: could we check for a pk in save? and only write this value on create?

codalab/apps/web/models.py
```diff
@@ -351,7 +359,7 @@ def get_chahub_data(self):
             submitted_at__gt=now() - datetime.timedelta(days=30)
         ).exists()
 
-        return [{
+        return [self.clean_private_data({
```
jimmykodes: should be sending creator_id, too, not just username. probably some other stuff in here that may have broken w/ the adjustments to the endpoint on chahub, bbuuut I think the bulk (and important stuff) should still be working...

codalab/apps/web/models.py
```diff
@@ -2350,6 +2366,34 @@ def write_multidataset_metadata(self, datasets=None):
 
         self.data_file.save("metadata", ContentFile("\n".join(lines)))
 
+    def get_absolute_url(self):
```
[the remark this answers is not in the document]
ckcollab: ahhh I could go either way but I'll switcheroo to `get_download_url` to match v2 implementation

codalab/apps/web/models.py
```diff
@@ -316,6 +316,14 @@ def set_owner(self, user):
     def get_chahub_endpoint(self):
         return "competitions/"
 
+    def get_whitelist(self):
+        return [
+            'remote_id',
+            'created_by',
```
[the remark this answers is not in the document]
ckcollab: yeah I want that PK to exist still so connections still work, I think that's what we want? the profile attached to this user's PK would be where private data should be scrubbed

codalab/apps/web/models.py
ckcollab:
```
'creator_id': self.uploaded_by.id,
```

am I missing something or that's already being sent? the username is _also_ attached but maybe not super useful?

I tested sending a dataset again, appears to still work with one slight hiccup I'll PM you about

codalab/apps/web/models.py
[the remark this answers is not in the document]
ckcollab: ✔️

codalab/apps/web/models.py
ckcollab: added check for PK

codalab/apps/web/models.py
jimmykodes: This is a competition, not dataset, I don't see creator_id here.

codalab/apps/web/models.py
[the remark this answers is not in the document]
jimmykodes: this isn't a PK, though, it's a username string.

codalab/apps/web/models.py
ckcollab: my mistake. fixed.

jimmykodes: uh, sorry, created_by _is_ a string, `creator_id` is an additional field, that should be a pk, and the one we whitelist.

codalab/apps/web/models.py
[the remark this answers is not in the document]
ckcollab: another mixup `Competition` vs `OrganizerDataset`, it's a PK now

jimmykodes:
Should we also send submissions proper? or just their dataset?

codalab/apps/web/models.py
ckcollab: To clarify, for Competitions I'll whitelist `creator_id` and send along the following fields to Chahub:

```            "created_by": str(self.creator),
            "creator_id": self.creator.pk,```

(confirmed from v2 this is how you're doing it)

**Comment by ckcollab** [comment]
Submissions are already sent, but we may want to whitelist/privatize some of the data?

```
    def get_chahub_data(self):
        return {
            "remote_id": self.id,
            "competition": self.phase.competition_id,
            "phase_index": self.phase.phasenumber,
            "participant": self.participant.user.username,
            "submitted_at": self.submitted_at.isoformat(),
        }
```

whitelist everything _but_ participant username?

Should probably just be attaching participant user id anyway ??

PR #2731, opened 2019-12-14
charepo in v1.8

ckcollab:


ckcollab:
Woo! Able to publish competitions/datasets now, used AutoDL staging server to test this, seems to be working O K

codalab/apps/chahub/tasks.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+@task(queue='site-worker')
+def do_chahub_retries(limit=None):
+    if not chahub_is_up():
+        return
+    chahub_models = get_chahub_models()
+    logger.info('Retrying for ChaHub models: {}'.format(chahub_models))
+
+    chahub_models = inheritors(ChaHubSaveMixin)
```
jimmykodes: use `get_chahub_models`?

codalab/apps/chahub/tasks.py
```diff
@@ -0,0 +1,221 @@
...
+        # then delete all objects that need to be deleted
+        objects_to_be_deleted = model.objects.all_objects().filter(
+            chahub_timestamp__isnull=False,
+            chahub_needs_retry=False,
+            deleted=True
+        )
+        if limit is not None:
+            objects_to_be_deleted = objects_to_be_deleted[:limit]
+        for obj in objects_to_be_deleted:
+            obj.delete(real_delete=True)
```
jimmykodes: Why do we force real delete here? shouldn't we `real_delete` only on success from chahub?

codalab/apps/chahub/tasks.py
```diff
@@ -0,0 +1,221 @@
...
+def send_everything_to_chahub(limit=None):
+    if not chahub_is_up():
+        return
+    chahub_models = get_chahub_models()
+    for model in chahub_models:
+        batch_send_to_chahub(model, limit=limit)
+
+
+@task(queue='site-worker')
+def send_chahub_general_stats():
```
jimmykodes: Pretty sure none of this stuff here matters, since these statistics are calculated based on objects in the db on chahub, not as static counts anymore.... unless that isn't how we want chahub to work.

codalab/apps/chahub/tasks.py
```diff
@@ -0,0 +1,221 @@
...
+
+def batch_send_to_chahub(model, limit=None, retry_only=False):
+    qs = model.objects.all()
+    if retry_only:
+        qs = qs.filter(chahub_needs_retry=True)
+    if limit is not None:
+        qs = qs[:limit]
+
+    endpoint = model.get_chahub_endpoint()
+    data = [obj.get_chahub_data() for obj in qs if obj.get_chahub_is_valid()]
```
jimmykodes: because `get_chahub_data` returns `[{data:data}]` this will cause this array to look like: `[[{data: data}], [{data: data}]]` which will not be readable by chahub. (at least, this looks to be the case on competitions and datasets)

codalab/apps/chahub/models.py
```diff
@@ -119,5 +155,19 @@ def save(self, force_to_chahub=False, *args, **kwargs):
...
+                self.chahub_needs_retry = False
+                super(ChaHubSaveMixin, self).save(force_update=True)
+
+    def delete(self, real_delete=False, **kwargs):
+        if real_delete:
+            super(ChaHubSaveMixin, self).delete(**kwargs)
+        else:
+            # So this can be sent to ChaHub, with retries and such, don't actually delete
+            self.deleted = True
+            self.save()
```
jimmykodes: we want to _not_ send to chahub here, right?

codalab/apps/chahub/tasks.py
```diff
@@ -0,0 +1,221 @@
...
+    headers = {'X-CHAHUB-API-KEY': settings.CHAHUB_API_KEY}
+
+    try:
+        resp = requests.delete(url=url, headers=headers)
+    except requests.exceptions.RequestException:
+        resp = None
+
+    if resp and resp.status_code == 204:
+        logger.info("ChaHub :: Received response {} {}".format(resp.status_code, resp.content))
+        obj.delete()
```
jimmykodes: This is where we wanna pass `real_delete`?

ckcollab: added

codalab/apps/chahub/tasks.py
ckcollab: that's true, how is v2 handling this? copied that code from there

codalab/apps/chahub/tasks.py
ckcollab: believe i had to use this way because of python2 and lower versions of Django. in 1.8 I might be able to? would rather not have to confirm that and stick with this, which should be working

codalab/apps/chahub/tasks.py
ckcollab: because these objects were sent to chahub successfully, don't need to be retried, and were deleted -- I _think_ that makes sense...

codalab/apps/chahub/tasks.py
ckcollab: cool will pull

codalab/apps/chahub/models.py
ckcollab: I believe we do? can hop in a chat to talk about this stuff if that's quicker

jimmykodes: don't need to send to chahub here since the `delete_from_chahub` below is sending the same information again

ckcollab: Adding a kwarg to save so we can `send_to_chahub=False`

codalab/apps/chahub/tasks.py
jimmykodes: remove array from each `get_chahub_data` and wrap it in a list in the save method

codalab/apps/chahub/tasks.py
```diff
@@ -0,0 +1,194 @@
...
+            objects_to_be_deleted = objects_to_be_deleted[:limit]
+        for obj in objects_to_be_deleted:
+            obj.delete(real_delete=True)
+
+
+@task(queue='site-worker')
+def send_everything_to_chahub(limit=None):
+    if not chahub_is_up():
+        return
+    chahub_models = get_chahub_models()
```
jimmykodes: Check validity of this

codalab/apps/chahub/tasks.py
jimmykodes: call celery task directly here?

codalab/apps/chahub/tasks.py
ckcollab: on closer inspection: both were calling `inheritors`?

```
def get_chahub_models():
    return inheritors(ChaHubSaveMixin)
```

codalab/apps/chahub/models.py
ckcollab: ✔️

codalab/apps/chahub/tasks.py
ckcollab: ✔️

codalab/apps/chahub/tasks.py
ckcollab: ✔️

codalab/apps/chahub/tasks.py
ckcollab: ✔️

codalab/apps/chahub/tasks.py
ckcollab: we were doing `get_chahub_models` above anyway already, I'm getting tired :(

ckcollab:
Staging server: http://34.74.117.238/

PR #2735, opened 2020-01-14
First modification of header in index.html 

Didayolo:
* Delete a part of index.html header.
* Conversion of favicon.ico to real ico format.


ckcollab:
We should be making PRs into `develop` -- is there a reason this is directly to master?

I believe we have to fix a front page issue on `develop` as well, before we can deploy this.

Didayolo:
You are right, for some reason I thought the front page message was not on `develop` branch (it got merged with `lri-master` and has the "academic" front page.

Didayolo:
I think I'll re-create this PR from `develop` branch but you should be aware that now on develop we have the front page from the academic version of Codalab...

PR #2737, opened 2020-01-20
Remove crash message.

Didayolo:


Didayolo:
I can confirm that the `index.html` webpage (with the "Dear Codalab users" message) doesn't appear on `develop` branch version of Codalab. 

The frontpage is the one with "LRI students", "Wizard", etc.

developcodalab


This does not seem right.

Issue #2758, opened 2020-03-30
[Feature request] Weighted average ranking on multi-tasks leaderboard


Didayolo:
We can rank participants with the average ranking on several scores using something like this:

```
leaderboard:
  columns:
    auc_binary:
      label: Classification (AUC ROC)
      leaderboard: &id001
        label: RESULTS
        rank: 1
      rank: 2
      sort: desc
    auc_binary_2:
      label: Feature Selection (AUC ROC)
      leaderboard: *id001
      rank: 3
      sort: desc
    ave_score:
      label: < Rank >
      leaderboard: *id001
      rank: 1
      sort: asc
      computed:
        operation: Avg
        fields: auc_binary, auc_binary_2
  leaderboards:
    RESULTS: *id001
```

But some organizers would like to be able to be able to compute an weighted average ranking, let's say: **Task_A_rank * 0.4 + Task_B_rank * 0.6**

We can see in this piece of code from Codalab that the current operation implemented are average and maximum:

```
class SubmissionComputedScore(models.Model):
    scoredef = models.OneToOneField(SubmissionScoreDef, related_name='computed_score')
    operation = models.CharField(max_length=10, choices=(('Max', 'Max'),
                                                        ('Avg', 'Average')))
```

The coefficient in the weighted average would represent the importance of each sub-task.

Didayolo:
Feature implemented in `develop` branch.

PR #2762, opened 2020-03-31
Implement weighted average rank computation.

Didayolo:
Feature requested in the following issue: #2758 

It basically consists in compute the average rank on a multi-task competition, with coefficients defined in `competition.yaml` file like this:

```
ave_score:
    label: < Rank >
    leaderboard: *id001
    rank: 1
    sort: asc
    computed:
        operation: Avg
        fields: auc_binary, auc_binary_2
        weights: 0.8, 0.2
```

Then it computes `score1_rank * 0.8 + score2_rank * 0.2`. 

/!\ For ~~descending~~ ascending rank (lower is better), small coefficients means a more important task. /!\

Here is a screenshot of the results as excepted:

leaderboard_weighted_average

You can see here the normal results without `weights` field in the `competition.yaml` file:

leaderboard_weighted_2

# Tested
- [X] competition with a weighted average rank
- [X] competition with average rank not broken

# Checklist
- [X] Code review by me 
- [X] Hand tested by me 
- [X] I'm proud of my work
- [x] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

# TODO
- [ ] Documentation on Wiki (@Didayolo) 
- [ ] Add the field in EDIT menu ?
- [x] Fix weights not being nullable (@ckcollab)
- [x] Add test for weights when generating scores (@ckcollab)
- [x] Raise errors when unpacking invalid weights (@ckcollab)


ckcollab:
Merged develop in, should fix tests.

Would you like me to whip up a quick tests for this weights stuff?

Didayolo:
> Would you like me to whip up a quick test for this weights stuff?

Yes please, this would be awesome!

codalab/apps/web/models.py
```diff
@@ -2213,6 +2225,7 @@ class SubmissionComputedScore(models.Model):
     scoredef = models.OneToOneField(SubmissionScoreDef, related_name='computed_score')
     operation = models.CharField(max_length=10, choices=(('Max', 'Max'),
                                                         ('Avg', 'Average')))
+    weights = models.CharField(max_length=200)
```
ckcollab, line 2228: should be nullable, doing this

ckcollab:
Adding this to edit form seems like it will be complicated, should that be a blocker to merging this?

Didayolo:
> Adding this to edit form seems like it will be complicated, should that be a blocker to merging this?

I don't think so. The leaderboard rules are already hard to edit from this menu, and this feature won't be used a lot in my opinion.

ckcollab:
Cool, to confirm, this is ready to ship?! 

Once merged into `develop` we should do a heavy round of testing. @Tthomas63 are you available soon to spin up latest `develop` locally and help us find outstanding bugs?

Didayolo:
Something seems strange to me in the test.

We have `self.score_def` and `self.score_def_2` using **descending order**. So to have: 
```
organizer rank = 1.2
participant rank = 1.8
```

The organizer should be worse than the participant on the first score (which is the case, 120 > 100, so with descending order it is indeed worse) and better on the second score; this time we have '-' vs '100'. 

Why do we have `assert organizer_score['values'][1]['val'] == '-'` ? And isn't the "organizer" loosing against the "participant" on this second score?

The value computed by the weighted average rank for one participant really depends on the scores obtained by other participant as it uses **ranking**.

ckcollab:
> Why do we have assert organizer_score['values'][1]['val'] == '-' ? And isn't the "organizer" loosing against the "participant" on this second score?

It looks like that's an empty score column for that submission. Do we need to modify the scoring code to account for this?

Didayolo:
> It looks like that's an empty score column for that submission. Do we need to modify the scoring code to account for this?

I don't think so, it is OK. 

Also, I've made a confusion between descending and ascending order. The test finally looks OK to me.

We have the weights:
```
weights = "0.8, 0.2"
```

And the following ranking:
```
organizer rank = 1.2
participant rank = 1.8
```

This means that the organizer won the first task and lost the second (ranked 1 * 0.8 + ranked 2 * 0.2). With a descending order score, it is the case (won with 120 > 0, lost by having no submission).

Issue #2948, opened 2021-05-12
Disable the admin feature "Re-run all submissions"


Didayolo:
In the admin "Submissions" panel, we have this re-run all option:

![issue_screenshot](https://user-images.githubusercontent.com/11784999/118012879-cf811c00-b351-11eb-9d8b-9676c3d34a64.png)

We should disable it, as it is not that useful and it can cause severe queue congestion when misused.

Didayolo:
Done.

PR #2961, opened 2021-06-15
Add a superuser or staff condition to rerun all submission button

loicsarrazin:
Feature requested from the following issue : #2948

The buttom is disabled for competition admins or owners, unless they have superuser/staff access

image
image

Note that the UI is not as the one in the original issue since it seems to have changed following commit debc04349ff7a3942edbf2d7dd8fbfe1f8141170

The change required adding a flag to the competition submissions page view on the user's access level, communicating it to the js script, and having the Vuetify button put conditional to its value.

# Checklist for hand testing
- [x] competition admin/owner with superuser and/or staff access can see and click to the button
- [x] competition admin/owner without superuser and/or staff access can't see the button
- [x] button is still re-launching every submissions

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/codalab/apps/web/templates/web/my/submissions.html
+++ b/codalab/apps/web/templates/web/my/submissions.html
@@ -181,7 +181,7 @@
                                     Select a phase to view details & run actions against it<br>
                                 </span>
 
-                                <v-btn
+                                <v-btn v-if="is_superuser_or_staff"
                                         :disabled="!phase"
                                         @click="re_run_all_submissions"
                                 >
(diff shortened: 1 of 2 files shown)


Didayolo:
Congrats for such a clean pull request. I see you have found a good solution for this.

Some remarks:
* In the "superuser" screenshot, the button appears to be grey. Is the button functional? Anyway, I need to hand-test this by myself.
* We should check if the Wiki need to be updated, at least to add this information somewhere.
* You mentioned that there may be a security issue. I think it's not too concerning but we should look quickly into this.

loicsarrazin:
> In the "superuser" screenshot, the button appears to be grey. Is the button functional? Anyway, I need to hand-test this by myself.  
  
The button needs a phase to be selected to be activated. None were selected in the screenshot

Issue #2970, opened 2021-07-16
[CodalabPy3] Competition's submission drop down menu selector not working in py3 version


loicsarrazin:
Using Google Chrome
Platform is locally hosted

In the comptetition's submissions page, the submmission drop down selector does not work on the python 3 version.

Below is an example of one of the sample competition created with a submission, under both versions.

Using lri-py3-storage branch
![No Filter](https://user-images.githubusercontent.com/83833966/125979677-1d672a73-d1b6-4279-94f0-e5e2c8539c42.png)
![With Phase 3 filter](https://user-images.githubusercontent.com/83833966/125979717-46a49ea2-4536-4600-bbd8-f619454441ca.png)

Using develop branch
![No Filter](https://user-images.githubusercontent.com/83833966/125980527-9b41bb0d-ceb1-4435-a5e4-e092bd9ffa79.png)
![With Phase 3 filter](https://user-images.githubusercontent.com/83833966/125980789-ddf67a06-e9e6-44c1-9806-3c923ebdcf7e.png)

PR #2972, opened 2021-07-20
submissions page filter to check for equal phase's phasenumber and su…

loicsarrazin:
 Issues on the matter : https://github.com/codalab/codalab-competitions/issues/2970 
 
Small fix so that the filtering on phases works for the submission list in the competition admin's submission page.

Before the fix, the filter would have no impact, since it was trying to compare non-existing attributes of phases and submissions
In this example prior to the fix, the submission's phase is not in line with the one selected in the filter
image  

For the same example after the fix, now the submission is excluded from the list
image

- [x] Leaving no selection shows all submissions
- [x] Selecting a phase filters out all submissions not for this phase
- [x] Having submissions for different phases doesn't impact the behavior

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/codalab/apps/web/templates/web/my/submissions.html
+++ b/codalab/apps/web/templates/web/my/submissions.html
@@ -472,7 +472,7 @@
                         filtered_submissions() {
                             return this.submissions.filter(s => {
                                 if (this.phase) {
-                                    return this.phase.id !== s.phase.id
+                                    return this.phase.phasenumber == s.phase_number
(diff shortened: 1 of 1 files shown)


Didayolo:
Thanks for the clean PR ;-)

PR #2973, opened 2021-07-22
py3 Submissions page filter fix

loicsarrazin:
this is the py3 version of this PR : https://github.com/codalab/codalab-competitions/pull/2972

 Issues on the matter : https://github.com/codalab/codalab-competitions/issues/2970 
 
Small fix so that the filtering on phases works for the submission list in the competition admin's submission page.

Before the fix, the filter would have no impact, since it was trying to compare non-existing attributes of phases and submissions
In this example prior to the fix, the submission's phase is not in line with the one selected in the filter
image  

For the same example after the fix, now the submission is excluded from the list
image

- [ ] Leaving no selection shows all submissions
- [ ] Selecting a phase filters out all submissions not for this phase
- [ ] Having submissions for different phases doesn't impact the behavior

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/codalab/apps/web/templates/web/my/submissions.html
+++ b/codalab/apps/web/templates/web/my/submissions.html
@@ -180,7 +180,7 @@
                                     Select a phase to view details & run actions against it<br>
                                 </span>
 
-                                <v-btn
+                                <v-btn v-if="is_superuser_or_staff" 
                                         :disabled="!phase"
                                         @click="re_run_all_submissions"
                                 >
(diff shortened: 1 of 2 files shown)


Didayolo:
As mentioned in its comments section (#2961), it seems that the commits of this PR are exactly the same as the ones on this PR: #2961

Thus, it seems that we don't need to merge this one.
Do you agree?

loicsarrazin:
Yes. Let's close it,. Thank you!

PR #3032, opened 2021-11-16
Develop -> Master

Tthomas63:
Update master branch


Didayolo:
@Tthomas63 We basically want to replace the conflicting files by the `develop` version, right?

Tthomas63:
@Didayolo Not always, but for most of everything in this case yes. There was some differences on the highlights page, and I left the master branch's version of the different URL's commented out. That page is going to see work soon regardless

Didayolo:
@Tthomas63 Great work! Indeed I've seen the changes you are referring too. 

I guess this need hand-testing before merging. I'll try out soon.

loicsarrazin:
After verifying https://github.com/codalab/codalab-competitions/pull/3032/commits/05f79831b84a6b3b5af681ae6f137d74811cb179, that should be the only differences to develop, are not breaking the impacted views. I'm ok with merging this PR.

Didayolo:
I've also checked the code and tested manually, and everything seems good to go!

PR #3035, opened 2021-11-17
Add new URLs to admin monitoring links page.

Didayolo:
# Update admin monitoring page

I've added health status, storage analytics, etc. with descriptions and screenshots.
I have currently an issue with my local instance, so I've pushed in order to save the work but everything still needs to be tested.

- [ ] Log in with an admin account (superuser / staff)
- [ ] Click on "admin monitoring links" under the profile
- [ ] Check all buttons

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


codalab/apps/web/templates/web/admin_monitoring_links.html
```diff
@@ -22,4 +22,36 @@ <h3>Flower</h3>
             <p>Flower is a powerful web based Celery monitoring tool designed to keep track of our tasks. Admins may view the state of which tasks were run, with what arguments, and many more features. Here you may also view which queues your celery workers are consuming, and the state of any tasks in them. At last, there is also a great monitoring page for viewing the systemic impact of your workers.</p>
         </div>
     </div>
-{% endblock %}
\ No newline at end of file
+    <div class="row">
+        <div class="col-md-5" style="background-color: rgba(0, 0, 0, 0.05); padding: 20px;">
+            <a href="//{{ domain }}/health/simple_status/" target="_blank">
```
Tthomas63, line 27: If we change this to use url tags (`{% url %}`) it will always work, even if we change the actual url later. 

IE: I think `{% url "health:health_status_simple" %}` will create the proper href. The format is usually `<app_name>:<url_name>`

Didayolo: Hey, thanks for the review.

This is a great suggestion! Also, it makes the URLs work even if the domain name is ill defined (example.com).

I've made the change for health simple_status, status and storage analytics. Is there any similar URL to access Django admin console?

Didayolo: It works! I've added it.

Issue #3034, opened 2021-11-17
Submission status is "running" for a long time and all submission view is empty


nyokoya:
I am the organizer of the following competition:
https://codalab.lisn.upsaclay.fr/competitions/234?secret_key=631417b0-1962-4154-8f32-2a469713d52b

I have the following two issues:
1) Submission status is "running" for a long time for participants but I (organizer) cannot reproduce this issue.
2) All submission view is empty.

I have already added `pavao` and `tthomas63` as admins to my competition.
Many thanks for your help!

dalision:
the same issue

Didayolo:
Hi,

1. About the submission status, we will check the queue because it seems to be a general problem. Maybe the job queue is flooded.

2. About the "all submissions" view, there is another problem related (#3027). @loicsarrazin maybe you could have a look at this when you have time? As you have already made contributions to this page (#3031).

loicsarrazin:
Hi @nyokoya, would you please add me as a competition admin? My username is loic.sarrazin.  
  
 Thank you!

nyokoya:
@loicsarrazin I wonder if you could have time to take a look at the issue mentioned above

Didayolo:
@loicsarrazin is actively looking into it

loicsarrazin:
@nyokoya I think there might be a problem with the actual scoring, I suggest you change your scoring program's score with something else than "Prediction:None", because I suspect there is an error with the submission, and then a problem on our side with the parsing of a "None" results.  
  
 You can see in this example https://github.com/codalab/competition-examples/blob/master/basic-competition-bundles/Iris/iris_competition_bundle/scoring_program/score.py that "-1" is used.  
   
  I'll be giving updates on wether there is a problem on our side and when it's corrected

Didayolo:
About this side of the bug: "All submission view is empty", here are the error logs:

```
django            | 2021-11-25 16:03:35,866 INFO Calculating sub size for submission: 2454
django            | 2021-11-25 16:03:35,985 INFO Calculating sub size for submission: 2447
django            | 2021-11-25 16:03:36,080 INFO Calculating sub size for submission: 2446
django            | 2021-11-25 16:03:36,172 INFO Calculating sub size for submission: 2146
django            | 2021-11-25 16:03:36,181 ERROR Internal Server Error: /api/competition/234/submissions/
django            | Traceback (most recent call last):
[...]

PR #3043, opened 2021-11-26
testing the existence of a file in storage when accessing its size

loicsarrazin:
# @ mention of reviewers
@Tthomas63  I would like your advice on this one. 

In https://github.com/codalab/codalab-competitions/issues/3034, a user experienced failures of some submission size requests, which broke the all submissions views and the user's submission view. After investigating, it turns out that a submission has an object and a file related to it, but it is nowhere to be found on the storage, therefor crashing when requesting its size in the size computation.

# Known issues to be addressed in a separate PR
We need to figure out how such submission was created and why did it fail to be properly uploaded to the storage.

- [x] create a new submission
- [x] alter s3 key of submission so that it looks for an non-existing file
- [x] see that the submission view loads still
- [x] observe log of error

# Misc. comments
I'm not sure if there are security issues with the display of a bucket name in logs?  
Also if you have any idea on how to figure out how this happened let me know

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/codalab/apps/web/utils.py
+++ b/codalab/apps/web/utils.py
@@ -1,4 +1,5 @@
 import logging
+import traceback
 import os
 import re
 import requests
@@ -255,6 +256,8 @@ def get_size_from_summary(bucket_name, key):
         else:
             # Something else has gone wrong.
             raise
+    except Exception as e:
+        logger.error(traceback.format_exc())
     return size
 
 def delete_key_from_storage(obj, attr, aws_attr=None, s3direct=False, use_boto_method=True):
--- a/codalab/apps/web/utils.py
+++ b/codalab/apps/web/utils.py
@@ -255,7 +255,7 @@ def get_size_from_summary(bucket_name, key):
             logger.error("File was not found in storage. File: {0}; Storage: {1}".format(bucket_name, key))
         else:
             # Something else has gone wrong.
-            raise
+            logger.error(traceback.format_exc())
     except Exception as e:
         logger.error(traceback.format_exc())
     return size
--- a/codalab/apps/web/utils.py
+++ b/codalab/apps/web/utils.py
@@ -3,6 +3,7 @@ import os
 import re
 import requests
 import boto3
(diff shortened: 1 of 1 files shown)


codalab/apps/web/utils.py
```diff
@@ -238,12 +237,26 @@ def get_filefield_size(obj, attr, aws_attr=None, s3direct=False):
...
+        obj_summary = s3.ObjectSummary(bucket_name, key)
+        if obj_summary:
+            size = obj_summary.size
+    except ClientError as e:
+        if e.response['Error']['Code'] == "404":
+            # The object does not exist.
+            logger.error("File was not found in storage. File: {0}; Storage: {1}".format(bucket_name, key))
+        else:
+            # Something else has gone wrong.
+            raise
```
Didayolo, line 257: I guess we are interested in catching 404 error (when the object does not exist). But is this `raise` syntax correct? Will it provide the complete back traces if called? Also, isn't the whole point of this function to return None if a bug is encountered?

Maybe we should just remove lines 255 - 257

loicsarrazin: Indeed we are trying to catch a Client error which has an error code 404. But you are right that in case we catch a Client error with another error code, the trace back won't be informative. I'll apply some change to the code.

Didayolo: Yes I get it but I think we should return None if an error is encountered, hence removing lines 255 - 257.

If we raise an error here, this means that the views "My submissions" and "All submissions" will bug.

Didayolo: I approve your change since you still get the full traceback, thus debugging other cases remain possible.

Didayolo: Shouldn't you also replace `raise` by `logger.error(traceback.format_exc())` ?

PR #3065, opened 2021-12-16
Disabling competition creation button

loicsarrazin:
Commenting out competition creation


Didayolo:
The old production server (competitions.codalab.org) is actually running `archive-master-py2` branch (or more precisely an old version of the `master` branch).

But anyway, we'll just perform the change by hand.

loicsarrazin:
Oh, my bad.  
We can change the target branch if you want it to be in the repo, or just close this PR as it mostly was meant to highlight the change to be made.

PR #3133, opened 2022-03-28
Add github workflow file for running pytest

Tthomas63:
Add github workflow for running pytest

### DIFF
--- a/.github/workflows/run-pytest.yaml
+++ b/.github/workflows/run-pytest.yaml
@@ -50,6 +50,7 @@ jobs:
             ${{ runner.os }}-
       - name: Install dependencies
         run: |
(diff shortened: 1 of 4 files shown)


bbearce:
I see a discussin [here](https://github.com/jazzband/django-configurations/issues/236) on this issue and I'm finding it a little hard to understand what needs to be done. There is also a discussion linked from the previous link in the last sentence [here](https://github.com/typeddjango/django-stubs/pull/180). I'm not really sure what they are talking about at the moment. Could this be because we aren't really deploying inside a docker container, but rather a VM?

Didayolo:
@xbaro 
This is the pull request with your changes regarding test coverage.

PR #3134, opened 2022-03-28
highlights page stats to match codabench

bbearce:
# @ mention of reviewers
@Didayolo 

Add stats for:
* Total Competitions
* Public Competitions
* Private Competitions
* Users
* Competition Participants
* Submissions

- [x] Create User
- [x] Create Competition A (published or "public")
- [x] Create Competition B (Unpublished or "private")
- [x] Create Submission

# Preview
image

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/codalab/apps/web/templates/web/highlights.html
+++ b/codalab/apps/web/templates/web/highlights.html
@@ -13,6 +13,9 @@
             href="https://cdn.jsdelivr.net/semantic-ui/2.2.10/semantic.min.css">
         <script src="https://cdn.jsdelivr.net/semantic-ui/2.2.10/semantic.min.js"></script>
         <style>
+            #general-stats {
+              margin: 3em 0 0 0;
+            }
             p {
             text-align: justify !important;
             word-wrap: break-word !important;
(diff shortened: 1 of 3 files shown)


bbearce:
Now that I remember, we may have talked about adding this to the front page. I can do that pretty easily and it just slipped my mind. Let me know if that is still preferred and I can do that.

bbearce:
Example deploy here: https://codalab-python3-test.eastus.cloudapp.azure.com/highlights

bbearce:
The spacing is an issue. I will look into that. 

Also wanted to note in Codabench the django model Q filter is:
...
            published_comps=Count('pk', filter=Q(published=True)),
            unpublished_comps=Count('pk', filter=Q(published=False)),
...
but in this PR it is:
...
            published_comps=Count(Case(When(published=True, then=1))),
            unpublished_comps=Count(Case(When(published=False, then=1)))
...
The reason is that using the Codabench query I was getting unpublished results counted in with published results. In other words published_comps was always counting unpublished comps and had the same value. This fixed it.

Didayolo:
> The reason is that using the Codabench query I was getting unpublished results counted in with published results. In other words published_comps was always counting unpublished comps and had the same value. This fixed it.

Nice catch. Do you think then there is a problem on CodaBench side?

bbearce:
The css for codabench and codalab is such that if the "Users" count is small the information is left aligned inside it's div. Since we aren't explicitly setting the widths of the divs in the stats section we can't use "margin: auto" to center. The reason we don't see this for other divs in the stats section is because they all have long description text. By changing "Users" to "CodaLab Site Users" we force the div to auto adjust spacing and position to create a centered look. 

Anyone who is better at css could do better but this is pretty simple and clean to me.

Didayolo:
@bbearce 

> Live deploy here: https://codalab-python3-test.eastus.cloudapp.azure.com/highlights

This time the change is not visible in your live deploy.

Didayolo:
That's great! I do a last round of testing and merge it

Didayolo (approved):

Didayolo:
The change is now live: https://codalab.lisn.upsaclay.fr/highlights

Thank you very much @bbearce !

PR #3154, opened 2022-04-14
Quick fix on the 'disable submissions' issue

OhMaley:

Quick fix on the 'disable submissions' issue.
Also this PR include a rollback on the 'read only' field on max submission size (it was not intended to be part of the last PR)

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/codalab/apps/chahub/tests/test_chahub_mixin.py
+++ b/codalab/apps/chahub/tests/test_chahub_mixin.py
@@ -1,6 +1,7 @@
 import datetime
 import mock
 from apps.authenz.models import ClUser
+from apps.customizer.models import Configuration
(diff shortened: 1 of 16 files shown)


Tthomas63:
Will test locally, but tests seem to be failing due to due `Configuration` model instance not existing in tests. Not sure if we're using pytest fixtures or django testcase setup methods for these specific tests, but we can make this a reusable util/pytest fixture if need be to avoid repeating the logic

Tthomas63 (approved):

PR #3205, opened 2022-08-02
Integrity crossorigin referrerpolicy

carandraug:
This PR has two commits:

1. ensure that it connects with https to cloudflare (use of `//` to keep the same protocol may make sense while keeping in the same domain but I don't think it does when linking elsewhere)
2. set the integrity, crossorigin, and referrerpolicy for links to external resources (except the ones with jsdelivr that don't use a specific version - but maybe they should lock to a specific version?)

There is also a bunch of `http://` use all over codalab that I could replace with `/` but I will wait for feedback for this PR first.


Didayolo:
@carandraug

Thank you for your efforts in making clear issues and pull requests tackling some problems of the platform.

We took some time to reply because most of the staff was on vacation during August. We will review your pull requests soon.

Tthomas63 (approved):

PR #3206, opened 2022-08-02
base template link to "/my/" instead of "/my" 

carandraug:
The header has links to `/my`, `/my/settings/`, `/accounts/password/reset/`, and `/accounts/logout/`. The first does not end in a `/` so when the user clicks on it, it goes through a 301 redirect to `/my/`. The others all end up in `/` but are generated by `url` in the template.

I'm not sure where that redirect is happening but I'm codalab setting up behind a proxy and the redirects breaks. This is probably an issue with my proxy configuration but using `/my/` fixes the issue while making it consistent with the other links in the header.

### DIFF
--- a/codalab/apps/web/templates/account/login.html
+++ b/codalab/apps/web/templates/account/login.html
@@ -63,7 +63,7 @@
             <button class="btn btn-primary margin-top" type="submit">{% trans "Sign In" %}</button>
         </form>
         <p>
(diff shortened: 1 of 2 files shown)


Didayolo:
@carandraug 

Thank you for your efforts in making clear issues and pull requests tackling some problems of the platform.

We took some time to reply because most of the staff was on vacation during August. We will review your pull requests soon.

Didayolo:
Reviewed and tested locally.

PR #3209, opened 2022-08-04
.env_sample: remove mention of USER_SWITCH_MIDDLEWARE unused since 33f29205e4e2e2d78

carandraug:
The environment variable `USER_SWITCH_MIDDLEWARE` seems to no longer have any effect for over 4 years (since 33f29205e4e2e2d78 when `userswitch.middleware.UserSwitchMiddleware` was replaced with `django_switchuser.middleware.SuStateMiddleware`) but it's mentioned on `.env_sample`. This commit removes it.

### DIFF
--- a/.env_sample
+++ b/.env_sample
@@ -113,13 +113,6 @@ SSL_PORT=443
 # Allowed hosts separated by space
 SSL_ALLOWED_HOSTS=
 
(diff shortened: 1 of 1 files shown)


Didayolo:
I confirm that the "switch user" admin functionality still works after this change.

Issue #3185, opened 2022-06-16
Additional features for the admin page about the upper-bound limit on competitions submissions


OhMaley:
This ticket lists all feature we want to add to the PR https://github.com/codalab/codalab-competitions/pull/3163 after it is merged.

- Move the default limit into the settings (so it can be easily configurable)
- Display the limit at the top of the admin page
- In the table add columns for
  - the effective limit
  - organizer email
  - competition link
  - button to force the effective limit to be equal to the maximum limit that the admin has set
  - Fix the redirection page
- Make the table sortable
- Create a wiki page about this view and add a question mark button somewhere on the admin page

PR #3266, opened 2022-09-29
Get rid of django-extra-view. Redesign admin-competitions-manager to use Vue.js and API routes

OhMaley:
# Description

This PR redesigns the admin-competitions-manager page in order to get rid of the django-extra-view library that was causing some issue.

# Content

The admin-competitions-manager page is accessible to admins at _/admin_competitions_manager/_

It contains a table listing all competitions (paginated).
You can edit the _upper_bound_max_submission_size_ field of any competition by clicking on its value.
If you want to edit this field for multiple competitions at once you can use the checkboxes on the far-left column.
A _select all_ checkbox is also available. Be aware that it will only select all competitions displayed on the current page (because of pagination). If you want to select ALL competitions regardless of what is displayed you can chose to display all competitions then use the _select all_ checkbox.

Once the modification is done you can click on the _Save_ button to save them in the DB (bulk update in one transaction).

Features listed in the #3185 will be done in a separated PR

# Misc. comments

Tested locally

# Reviewers

@bbearce

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/codalab/apps/api/routers.py
+++ b/codalab/apps/api/routers.py
@@ -1,5 +1,6 @@
 from apps.api.views import competition_views as views
 from apps.api.views import storage_views as storage_views
+from apps.api.views import admin_views as admin_views
 from django.conf.urls import url
 from rest_framework import routers
 
@@ -40,4 +41,7 @@ urlpatterns += (
     # Storage Analytics
     url(r'^storage/analytics', storage_views.GetExistingStorageAnalytics.as_view(), name="existing_storage_analytics"),
     url(r'^storage/usage-history', storage_views.GetStorageUsageHistory.as_view(), name="storage_usage_history"),
+    # Admin
+    url(r'^admin/competitions/list', admin_views.GetCompetitions.as_view(), name="competitions"),
+    url(r'^admin/competitions/update', admin_views.UpdateCompetitions.as_view(), name="update_competitions"),
 )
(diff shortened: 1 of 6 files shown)


bbearce:
Noticing that when I set max submission size to 140MB and then upload a 148MB file, that it is not rejected.

image

image

OhMaley:
So this PR only limits the range of the `max_submission_size` value an organizer can set on a specific competition.
What is happening here:
- before updating the value of the `upper_bound_max_submission_size` was probably set to 0 or a higher value than 148MB (may be 300MB ?)
- Then you changed it to be lower (140MB). Note that is does not change the current value of `max_submission_size`. Only the upper bound limit of the range in which the organizer can set it.
- Now you have something weird: The upper bound limit is below the current `max_submission_size`. So that means a user can still upload a file whose size is below the `max_submission_size`.

In this PR, to fix the issue, the process was either:
- to send an email to the organizer asking to change the current value of `max_submission_size` for its competition
- change it manually in the DB and notify the organizer of the change

An enhancement is going to be introduced in #3270 allowing the admin to 'synchronize' the current `max_submission_size` to the value of `upper_bound_max_submission_size` which will solve this problem.

PR #3300, opened 2022-12-15
Update upload_backup.py

dtuantran:
Since Django 1.8, management commands now use argparse instead of optparse to parse command-line arguments passed to commands. This also means that the way to add custom arguments to commands has changed: instead of extending the option_list class list, you should now override the add_arguments() method and add arguments through argparse.add_argument(). See https://docs.djangoproject.com/en/1.8/releases/1.8/#extending-management-command-arguments-through-command-option-list

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [X] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
       DJANGO_SETTINGS_MODULE: codalab.settings
       DJANGO_CONFIGURATION: Dev
     docker:
-      - image: circleci/python:3.8.3
+      - image: cimg/python:3.8
     steps:
       - checkout
       - restore_cache:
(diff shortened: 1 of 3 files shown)


Didayolo:
@dtuantran 

The change of the PR seems pretty simple and straightforward as it is only 2 new lines of code.

On a side note, what is this new "Run Pytests / run-tests (pull_request)" test from Github? Was it setup recently?
I guess it is failing because of a problem of configuration of the test itself, and that the error is not related to your PR.

PR #3372, opened 2023-05-22
Fix installation of deprecated anyjson package

xbaro:
Downgrade setuptools before installing required package anyjson. This is a temporary patch to avoid errors due to the use of deprecated anyjson Python package


Didayolo:
@xbaro Thank you for your help on this!

